Issue #110, opened 2015-01-08
Update Sass to 3.4.x


parkr:
Ref https://github.com/jekyll/jekyll-sass-converter/issues/31 which stems from a GitHub Support message.

Would you please `bundle update sass` on Pages? alternatively, we can add 'sass' to `lib/github-pages.rb`. Whichever you'd like. :)

/cc @gjtorikian @benbalter

ajkochanowicz:
Thanks, @parkr

gjtorikian:
`bundle update sass` does nothing; `sass` on GitHub Pages is currently set to 3.4.9, which is the latest release. 

![sass](https://cloud.githubusercontent.com/assets/64050/5669567/540e5548-972d-11e4-9871-e1d9c868d1a4.png)

Did you mean, update `jekyll-sass-converter`? We're at 1.2, but 1.3 is out.

parkr:
Then you should be fine, @ajkochanowicz!

@gjtorikian Thoughts on having a separate segment of gems we report but don't lock externally?

gjtorikian:
@parkr You mean basically publishing the Gemfile.lock used by Pages? I think it's A Good Idea. Maybe extend *versions.json* to include that data? @benbalter Does that make sense?

Closing this out, since we are on the furthest Sass we can go. Please reopen if I'm missing something!

PR #111, opened 2015-01-08
Report on dependencies we don't wish to lock.

parkr:
Ref #110.

### DIFF
--- a/lib/github-pages.rb
+++ b/lib/github-pages.rb
@@ -35,9 +35,22 @@ class GitHubPages
   # Versions used by GitHub Pages, including github-pages gem and ruby version
   # Useful for programmatically querying for the current-running version
   def self.versions
-    gems.merge({
-      "github-pages" => VERSION.to_s,
-      "ruby"         => RUBY_VERSION
-    })
+    gems.merge deps_version_report
+  end
+
+  def self.deps_version_report
+    require 'html/pipeline/version'
+    require 'sass/version'
+    require 'safe_yaml/version'
+
+    {
+      "ruby" => RUBY_VERSION,
+
+      # Gem versions we're curious about
+      "github-pages"  => VERSION.to_s,
+      "html-pipeline" => HTML::Pipeline::VERSION,
+      "sass"          => Sass.version[:number],
+      "safe_yaml"     => SafeYAML::VERSION
+    }
   end
 end
--- a/bin/github-pages
+++ b/bin/github-pages
@@ -11,13 +11,15 @@ Mercenary.program(:"github-pages") do |p|
   p.command(:versions) do |c|
     c.syntax "versions"
     c.description "lists the current gem dependency versions"
+    c.option "all", "--all", "Report all versions, not just ones we lock to (cannot be used with --gemfile)"
     c.option "gemfile", "--gemfile", "use Gemfile formatting"
     c.action do |args, options|
       if options["gemfile"]
         GitHubPages.gems.each { |g,v| puts "gem '#{g}', '#{v}'" }
       else
         require 'terminal-table'
-        puts Terminal::Table.new :rows => GitHubPages.gems, :headings => ["Gem", "Version"]
+        versions = options['all'] ? GitHubPages.versions : GitHubPages.gems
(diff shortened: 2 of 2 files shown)


gjtorikian:
I do not mind merging this.

benbalter:
@parkr can you explain the use case a bit more? What problem exists if we don't do this? How does this resolve that? Why these gems and not others? Any downside?

parkr:
> can you explain the use case a bit more?

The use case for this pull request is to report via pages.github.com/versions/ the versions of dependencies not locked down by this gem directly, but via bundler in the Pages project itself. It's possible that, for example, a PATCH version of Sass is released with a fix, but the Pages workers are running a previous version. Having this would at least offer insight that we don't currently have, but would very much like.

XhmikosR:
Definitely worth having this for the reasons @parkr explains above.

benbalter:
From the Readme: 

> The goal of the GitHub Pages gem is to help GitHub Pages users bootstrap and maintain a Jekyll build environment that most closely matches the GitHub pages build environment. The GitHub Pages gem relies on explicit requirements shared between both users' computers and the build servers to ensure that the result of a user's local build is consistently also the result of the server's build.

Would the project's goals be better served by explicitly versioning those sub-dependencies? (which would then appear on `/versions/`? The goal of the Gem is not to create a pipeline to list what versions are used, but to minimize environmental differences, wherever possible.

XhmikosR:
It can certainly help spot issues caused by other gem version differences.

parkr:
> The goal of the Gem is not to create a pipeline to list what versions are used, but to minimize environmental differences, wherever possible.

I'd be down to version where it makes sense. Does it slow down Pages further, however?

Limiting this gem to a single ruby version wouldn't be very wise, as Ruby hasn't changed too too much so it'd be exceedingly rare that 2.1.4 to 2.1.5 would cause a bug for you.

Point of this PR is to share these versions so we have a means of debugging. Pages shouldn't be a :black_large_square: if it can be helped. Otherwise, it becomes too difficult and too opaque to use.

parkr:
I don't see the harm in this.

parkr:
Pretty please?

lib/github-pages.rb
```diff
@@ -35,9 +35,22 @@ def self.gems
...
+    require 'html/pipeline/version'
+    require 'sass/version'
+    require 'safe_yaml/version'
+
+    {
+      "ruby" => RUBY_VERSION,
+
+      # Gem versions we're curious about
+      "github-pages"  => VERSION.to_s,
+      "html-pipeline" => HTML::Pipeline::VERSION,
```
gjtorikian: My only question is, why is `html-pipeline` in here? That's not exposed to GitHub Pages as far as I can tell you.

parkr: The motivation for reporting on these is that as these dependencies may differ locally versus the server. HTML Pipeline changed a bunch between 1.8 and 1.9 and I remember having trouble because of this, but I didn't have the Gemfile.lock for Pages, so I was stuck.

gjtorikian: What I mean is: who is using HTML Pipeline?

parkr: Oh oh oh, sorry! Mentions and emoji do, I believe :-)

gjtorikian:
I don't really see a problem with this PR. It's a philosophical difference between these two statements:

> The goal of the GitHub Pages gem is to help GitHub Pages users bootstrap and maintain a Jekyll build environment that most closely matches the GitHub pages build environment...The goal of the Gem is not to create a pipeline to list what versions are used, but to minimize environmental differences, wherever possible.

I personally think that listing versions would minimize environmental differences. I also think listing versions would help people build an environment that matches the build environment we use.

I might judge this PR a bit more scrupulously if it:

1. Involved maintenance for us to keep up-to-date
2. Were delivered by someone other than @parkr 

I'm fine with it, though. Pages is a bit of a black box to users. If it can be a little transparent, that's great.

parkr:
> I'm fine with it, though. Pages is a bit of a black box to users. If it can be a little transparent, that's great.

Thank you so so much, @gjtorikian!!!!

PR #127, opened 2015-04-09
add note in CONTRIBUTING.md about testing

boneskull:


benbalter:
Thanks @boneskull!

boneskull:
(I probably should have mentioned that if I had known this, I wouldn't have spent an hour sparring with Bundler to figure it out.)

boneskull:
(My Ruby-fu is weak)

PR #128, opened 2015-04-12
upgrade Redcarpet to 3.2.3

boneskull:
- release 3.2.3 fixes a potential XSS security vulnerability, as documented [here](http://seclists.org/oss-sec/2015/q2/69), [here](https://hackerone.com/reports/46916), and [here](http://danlec.com/blog/bug-in-sundown-and-redcarpet).

### DIFF
--- a/lib/github-pages.rb
+++ b/lib/github-pages.rb
@@ -15,7 +15,7 @@ class GitHubPages
       "kramdown"              => "1.5.0",
       "maruku"                => "0.7.0",
       "rdiscount"             => "2.1.7",
-      "redcarpet"             => "3.1.2",
+      "redcarpet"             => "3.2.3",
       "RedCloth"              => "4.2.9",
 
       # Liquid


boneskull:
Fix in vmg/redcarpet@7e9ba24f3c166f71804df18ea92e42af0373a969

parkr:
:+1:

spraints:
cc #134.

@github/appsec - do you have a few minutes to check out the redcarpet update? Here's a [diff of vmg/redcarpet from v3.1.2 to v3.2.3](https://github.com/vmg/redcarpet/compare/v3.1.2...v3.2.3).

boneskull:
It's quite possible this breaks something as I haven't had a chance to test it thoroughly

benbalter:
> fixes a potential XSS security vulnerability

My understand is the XSS vulnerability would only affect untrusted user content (e.g., where you're using Markdown to sanitize user input). Given that Pages allows arbitrary HTML input to begin with (rendering the attack vector moot), are there other advantages of upgrading?

boneskull:
@benbalter 

> Given that Pages allows arbitrary HTML input to begin with (rendering the attack vector moot)

Not sure I understand what you're getting at here.  Can you please provide an example?

benbalter:
My understanding of the XSS vulnerability is that it allows a user who has the ability to modify the Markdown to potentially place malicious tags in the rendered HTML.

> When markdown is being presented as HTML, there seems to be a strange interaction between _ and @ that lets an attacker insert malicious tags.

That's a vulnerability when Markdown is used as a sanitization layer (e.g., in GitHub comments), but on GitHub Pages, any user that can modify the markdown can also simply supply the malicious tag directly as HTML. It is then my understanding that the security vulnerability doesn't affect GitHub Pages (and thus we'd want another reason to motivate the Gem bump).

boneskull:
OK, fair enough.  Would like to see some movement on #132 though.  :wink:

PR #131, opened 2015-04-17
Update Travis CI configuration.

XhmikosR:
Test Ruby 2.0 - 2.2.

Please check the comments in https://github.com/XhmikosR/pages-gem/commit/0938fad9347a5e92694ce4a48499400f10bece3e for the reasoning.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,22 @@
 language: ruby
-rvm: 2.1
+
+rvm:
+  - 2.2
+  - 2.1
+  - 2.0
+
 script: "script/cibuild"
+
 env:
   global:
-  - NOKOGIRI_USE_SYSTEM_LIBRARIES=true
+    - NOKOGIRI_USE_SYSTEM_LIBRARIES=true
 
 notifications:
   email: false
 
 cache: bundler
+
 sudo: false
+
+matrix:
+  fast_finish: true


parkr:
:+1:

gjtorikian:
Thanks for opening this!

We talked a bit about this internally. The goal of pages-gem is to match Pages in production as closely as possible, if not exactly. Right now, Pages in prod runs Ruby 2.1, so `rvm` should be scoped to just that version. 

Normally, I'd be comfortable saying something to the effect of "But patch versions rarely change things!", except for the fact [that that's not entirely true](https://byparker.com/blog/2014/ruby-2-2-0-time-parse-localtime-regression/).

It would also be great if the README were updated to make explicit note that we're running Ruby 2.1 (probably under the "What's versioned" section). Going forward we'll make sure to update the Travis file + README if we ever upgrade our rubies. /cc @github/pages 

Thanks again!

parkr:
> Right now, Pages in prod runs Ruby 2.1, so rvm should be scoped to just that version.

We talked about this in the commit. Unless this gem _only_ works under 2.1 (i.e. it cannot even be installed under any other version), then it makes sense to test it under the compatible versions. It's unreasonable to expect everyone to know to use Ruby 2.1.x if it's not enforced (not all people read the readme, sadly). If we say `required_ruby_version("=> 2.0")`, then it makes sense to spend the extra 70 seconds to test them just for peace-of-mind. :smile: 

Maybe a warning can be displayed at install-time if you're not using 2.1?

gjtorikian:
That the gem is installed/runs on Ruby >= 2.0 is one thing. That the gem is *explicitly tested* for 2.1 is something else entirely. It's about sending a message.

If people make PRs to bump dependencies, I think it's reasonable to assume that we test them under just 2.1, since that's what we're testing for production, too. I guess I just don't see or understand the value of showing test results for 2.2+. The scenario I want to avoid is suggesting that in production we are also matching non-2.1 versions, when that's not the case at all. 

> Maybe a warning can be displayed at install-time if you're not using 2.1?

Maybe. That runs the risk of confusing people, too ("Why does the gem say this? WHAT DO I DO!"). I'd rather be cautious and wait to see if setting expectations for 2.1 becomes a problem before going that route.

XhmikosR:
While I see the point of view from GutHub, please try and see these:

1. The gem is public
2. You already allow Ruby >= 2.0
3. Testing only takes a few seconds
4. When things broke for 1.9.3 no one noticed apart from me (and that's because of my other PR, which makes no sense to me why it's not merged yet; I don't want to get more off topic here)
5. The platforms tested have nothing to do with the Ruby version Pages use. It's just to make sure things work right for other systems too, since the gem is **public**

That said, it makes totally sense to have this merged to me. But, again, I don't work for GitHub :P

gjtorikian:
> When things broke for 1.9.3 no one noticed apart from me (and that's because of my other PR, which makes no sense to me why it's not merged yet; I don't want to get more off topic here)

And I appreciate that greatly. It was a mistake on our part to not check for this earlier.

> It's just to make sure things work right for other systems too, since the gem is public

Again, there seems to be some confusion about the purpose of this gem. The purpose of this gem is not to provide a bootstrapped way to run a Jekyll project locally. The purpose of the gem is to recreate a local environment that matches the environment running on Pages. There's a fine line between the two. 

The production Pages environment runs Ruby 2.1, so that's what we use for CI too.

XhmikosR:
> And I appreciate that greatly. It was a mistake on our part to not check for this earlier.
I mentioned it to see my point: I wouldn't have noticed it if it wasn't for the CI PR.

Again, this isn't a gem that you only use on GitHub servers. It's public, in order for people to bootstrap the environment GitHub pages run (I never said anything about Jekyll here). So, since you don't provide Ruby versions yourself, you can't expect everyone to run "2.1.0-github". Thus, a basic testing is more than important here.

So, you wanna mention in README.md that you only care about 2.1.0-github? You could that, sure. You want to restrict installation to 2.1.0-github only? That would be stupid IMO, since it works fine to other versions. Here comes this PR.

Anyway, please try to be more open minded about this; it's for this project's and community's own good. Personally, I'm aware how to solve such issues and I don't even use the gem (because it's using so old dependencies).

Cheers.

parkr:
> The purpose of the gem is to recreate a local environment that matches the environment running on Pages.

Let's lock down this gem to the exact version being run in production. If I'm not running the right Ruby, I should be able to use this gem.

PR #132, opened 2015-04-17
update liquid to 2.6.2

boneskull:
- experiencing Shopify/liquid#502 with 2.6.1; a fix for this issue was in 2.6.2

### DIFF
--- a/lib/github-pages.rb
+++ b/lib/github-pages.rb
@@ -19,7 +19,7 @@ class GitHubPages
       "RedCloth"              => "4.2.9",
 
       # Liquid
-      "liquid"                => "2.6.1",
+      "liquid"                => "2.6.2",
 
       # Highlighters
       "pygments.rb"           => "0.6.1",


parkr:
:+1:

spraints:
LGTM

```
diff -r liquid-2.6.1/lib/liquid/htmltags.rb liquid-2.6.2/lib/liquid/htmltags.rb
46d45
<             'index0'  => index,
diff -r liquid-2.6.1/lib/liquid/version.rb liquid-2.6.2/lib/liquid/version.rb
3c3
<   VERSION = "2.6.1"
---
>   VERSION = "2.6.2"
```

parkr:
Thanks @spraints!! :smile:

benbalter:
Simply enough fix. [Full change log](https://github.com/Shopify/liquid/compare/v2.6.1...v2.6.2). Thanks for submitting @boneskull!

LGTM, but since it's just a warning level-message, don't believe we need to cut a new gem just for this.

parkr:
> LGTM, but since it's just a warning level-message, don't believe we need to cut a new gem just for this.

I disagree. The purpose of this gem is to make sure the versions everyone is using are exactly the same, right? I realize the diff is as surface-level as you can get, but this warning has been very noticeable so I think it makes sense for both reasons to cut a new :gem:.

boneskull:
@benbalter: It doesn't seem to be a warning; my site won't build w/o it.

PR #142, opened 2015-05-23
Updated everything.

KamranMackey:
Please, don't use outdated versions of things. That is a bad thing to do, especially when it comes to security issues & bugs. Oh, I also added the Rouge syntax highlighter to the list of highlighters for those who don't want to use Pygments.

### DIFF
--- a/lib/github-pages.rb
+++ b/lib/github-pages.rb
@@ -1,5 +1,5 @@
 class GitHubPages
-  VERSION = 35
+  VERSION = 36
 
   # Jekyll and related dependency versions as used by GitHub Pages.
   # For more information see:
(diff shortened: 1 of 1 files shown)


parkr:
@KamranMackey The Pages Team updates things incrementally and does so cautiously and expediently when a security vulnerability is discovered. Rest assured they're very good about security – that's why the updates aren't immediate too :)

KamranMackey:
Alright.  :smiley:

gjtorikian:
Yeah, we much prefer the PRs that look like this:

image 2015-07-29 at 4 49 40 pm

Thanks for caring, though. :heart: We are going to roll through that other list.

PR #149, opened 2015-06-29
Bumped Redcarpet to 3.3.2

Stargator:
Versions 3.3.2 has a fix for a potential security issue in the HTML renderer

### DIFF
--- a/lib/github-pages.rb
+++ b/lib/github-pages.rb
@@ -15,7 +15,7 @@ class GitHubPages
       "kramdown"              => "1.5.0",
       "maruku"                => "0.7.0",
       "rdiscount"             => "2.1.7",
-      "redcarpet"             => "3.3.1",
+      "redcarpet"             => "3.3.2",
       "RedCloth"              => "4.2.9",
 
       # Liquid


gjtorikian:
We actually deployed this internally to our Pages servers a week or so ago, as the security issue was disclosed to us by @vmg. Apologies that it's not in the public gem yet. 

Right now we are sussing out another potential issue, but we will merge and release this soon.

nerab:
Would be great to see this merged for the security fixes in redcarpet

Starefossen:
+1

gjtorikian:
As I said above:

> We actually deployed this internally to our Pages servers a week or so ago, as the security issue was disclosed to us by `@vmg`.

If you head to https://pages.github.com/versions/, you'll see the dynamically generated versions of all the gems we run in production. The entire company is heading out to our annual Summit next week, so we'll try to take care of this in the public release of the gem :soon:.

nerab:
github's pages are one thing, but dropcaster [depends on the gh_pages](https://github.com/nerab/dropcaster/blob/master/Gemfile.lock#L60), which still points to 3.3.1. So I cannot fix dropcaster until this PR was merged. Or is there a better way?

envygeeks:
@nerab Switch to the Github pages "like" Docker image that we provide on Jekyll then, we use almost the same dependency chain but like Github we diverge if we want to force a specific version for specific reasons.

gjtorikian:
> Or is there a better way?

I've just merged the PR in but am not doing a release just yet. Assuming you're using Bundler, you can refer directly to the gem with `:git => 'https://github.com/github/pages-gem.git', :branch => 'master'` until we can get a proper release out. Sorry for the trouble. Part of the complication here is that we were also [sussing out a segfault](https://github.com/github/pages-gem/issues/156) around the same time as the Redcarpet bump.

nerab:
OK, will try that. Thanks!

PR #151, opened 2015-06-29
Bumped Jemoji to 0.5.0

Stargator:
Prepares Jemoji to support Jekyll 3
Travis: add 2.2 to build matrix, remove 1.9.3, only master branch build


Stargator:
Contained unintended change, so removing this request and will replace it with the intended changes only.

PR #152, opened 2015-06-29
Bumped Jekyll-Sass-Converter to 1.3.0

Stargator:
Changes added by increasing the version number.

1.3.0
Include line number in syntax error message
Raise a Jekyll::Converters::Scss::SyntaxError instead of just a StandardError

1.2.1
Only include something in the sass load path if it exists


gjtorikian:
Seems like mostly improved error handling: https://github.com/jekyll/jekyll-sass-converter/compare/v1.2.0...v1.3.0

PR #159, opened 2015-06-30
Bundle health-check in github-pages.rb

Stargator:
So they can be easily updated in one file instead of having to update two files. The versions have remained the same as found in release v38

### DIFF
--- a/github-pages.gemspec
+++ b/github-pages.gemspec
@@ -18,7 +18,6 @@ Gem::Specification.new do |s|
     s.add_dependency(gem, "= #{version}")
   end
 
-  s.add_dependency('github-pages-health-check', "~> 0.2")
   s.add_dependency('mercenary', "~> 0.3")
   s.add_dependency('terminal-table', "~> 1.4")
   s.add_development_dependency("rspec", "~> 3.3")
--- a/lib/github-pages.rb
+++ b/lib/github-pages.rb
@@ -30,6 +30,7 @@ class GitHubPages
       "jekyll-redirect-from"  => "0.8.0",
       "jekyll-sitemap"        => "0.9.0",
       "jekyll-feed"           => "0.3.1",
+      "github-pages-health-check" => "0.3.1",
     }
   end
 


parkr:
I think we wanted to be more flexible about these, which is why they are in the gemspec with `~>` instead of `=`.

Stargator:
I saw that, but some of those went 2 years without a new version. And there's no telling whether that version would break anything in pages-gem.

rspec was ~> 2.14. a few days later 2.14.1 was release, but then 2.99.0 about 11 months later. And would fit the ~> 2.14 logic. If that broke anything, then changes would have to be made (either limit the version to the previous or update pages-gem).

While terminal-table is up to date. Between versions 1.5.0 and 1.5.1 there was a 4 year wait where who knows what it would have done when you didn't expect that change coming.

I understand your decisions. But the history of some of these dependencies would make me cautious about when the next change might come and what that could bring if a project doesn't have a bit more control over them.

gjtorikian:
Correct me if I'm wrong but one problem here is that `rspec` suddenly goes from a development dependency to a "real" dependency. That's no good. 

I do tend to agree with @parkr's logic here in that the gemspec files are items that the gem depends on, while the `self.gems` dependencies are files that Pages dependencies on. It's semantics, but I personally am in favor of keeping these separate.

gjtorikian:
(Didn't mean to close; I can be convinced either way given a strong enough argument. I just personally think keeping them separate is best.)

parkr:
> I just personally think keeping them separate is best.

:hand: me too.

Stargator:
@gjtorikian I updated the PR to resolve the issue that rspec was no longer a development dependency. I would like to ask, why you this "keeping them separate is best"?

I like the points I raised earlier. The most important being the large period of time some of the gems have waited for an update. If you depend on a gem that went 4 years before an update unexpectedly showed that. One that could still fit the ~> logic, but could conceivably introduce a breaking change.

It all just depends on the teams supporting those gems. Specifically, whether they are going to be active and cautious that they do not forget crucial details in the, potentially, years between releases.

If they were Github projects, then I wouldn't care. But it seems these are third-party dependencies and if they do happen to break something, then it's not just pages-gem, but also users that are affected.

From my point of view locking the versions down is better from a user's perspective since that reduces the likelihood of their experience being disrupted.

Either way, I just wanted to raise this point.

benbalter:
Going back to [the project goals](https://github.com/github/pages-gem#project-goals), the purpose of the Gem is to help users mirror the GitHub Pages production environment locally. That happens in two ways:

1. By specifically versioning gems used in production so the exact versions are used locally
2. By exposing the versions used in a machine-readable way, so they can be listed on https://pages.github.com/versions.

I'm fine moving health-check to being a versioned gem, since it's used in production, but believe versioning rspec, terminal table, et. al. which aren't used, and thus don't need to be in sync, would further confuse users. 

Put another way, I'm fine with it being slightly harder or slightly more confusing for us to maintain, if that means its ever so slightly easier for end-users to use.

parkr:
> I'm fine moving health-check to being a versioned gem, since it's used in production, but believe versioning rspec, terminal table, et. al. which aren't used, and thus don't need to be in sync, would further confuse users.

:+1:

Stargator:
@benbalter I'll make the update soon.

Stargator:
@benbalter Updated the PR. Let me know if health-check is fine where I placed it

benbalter:
Thanks @Stargator!

PR #162, opened 2015-07-01
Add Rouge support

jgoizueta:
Pygments is fine, but I like using fenced code blocks rather than highlight liquid tags. Fenced blocks are nicely highlighted by Atom in the source code, and can be rendered by kramdown+rouge using the same Pygment styles.

### DIFF
--- a/lib/github-pages.rb
+++ b/lib/github-pages.rb
@@ -1,5 +1,5 @@
 class GitHubPages
-  VERSION = 38
+  VERSION = 39
 
   # Jekyll and related dependency versions as used by GitHub Pages.
   # For more information see:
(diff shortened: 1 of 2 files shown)


thomasjo:
See https://github.com/github/pages-gem/pull/79.

jgoizueta:
OK, I guess I should look for a solution on the Jekyll/Markdown side that relies on Pygments for the time being.

Hope to have news about PrettyLights soon.

lib/github-pages.rb
```diff
@@ -1,5 +1,5 @@
 class GitHubPages
-  VERSION = 38
+  VERSION = 39
```
parkr: :question:

jgoizueta: Oops, I shouldn't have committed that

Stargator: You can close this pull request, create a new branch that only has the changes you desire and then do another pull request. I did that for a few requests that had unintended changes.

PR #169, opened 2015-08-12
Return the proper exit code from the DNS Check

benbalter:
This would allow users to call `github-pages health-check` as part of CI and receive the proper exit code.

### DIFF
--- a/bin/github-pages
+++ b/bin/github-pages
@@ -47,8 +47,10 @@ Mercenary.program(:"github-pages") do |p|
       puts "Checking domain #{cname}..."
       if check.valid?
         puts "Everything looks a-okay! :)"
+        exit 0
       else
         puts "Uh oh. Looks like something's fishy: #{check.reason}"
+        exit 1
       end
     end
   end


bin/github-pages
```diff
@@ -47,8 +47,10 @@ Mercenary.program(:"github-pages") do |p|
       puts "Checking domain #{cname}..."
       if check.valid?
         puts "Everything looks a-okay! :)"
+        exit 0
       else
         puts "Uh oh. Looks like something's fishy: #{check.reason}"
+        exit 1
```
parkr: You might consider `abort "Uh oh."` as a means of consolidating these two lines :)

bin/github-pages
```diff
@@ -47,8 +47,10 @@ Mercenary.program(:"github-pages") do |p|
       puts "Checking domain #{cname}..."
       if check.valid?
         puts "Everything looks a-okay! :)"
+        exit 0
```
parkr: Just for explicitity?

PR #177, opened 2015-11-03
Add jekyll last modified at plugin

foliosus:
I really want to be able to use the jekyll-last-modified-at plugin, so that I don't have to maintain those dates by hand in my pages. This PR adds it as a dependency to the project, using its most recent version.

Note: the second commit on this branch is the business. The first one is just formatting.

### DIFF
--- a/lib/github-pages.rb
+++ b/lib/github-pages.rb
@@ -7,29 +7,29 @@ class GitHubPages
   def self.gems
     {
       # Jekyll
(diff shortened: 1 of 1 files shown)


spraints:
Unfortunately, this plugin doesn't work on GitHub Pages. See gjtorikian/jekyll-last-modified-at#32 and #119.

PR #180, opened 2015-12-02
Bump minor gem versions in preparation for Jekyll 3.0.x

benbalter:
This pull request upgrades gems where there's a minor gem bump, in order to minimize the diff and number of moving parts in https://github.com/github/pages-gem/pulls.

### DIFF
--- a/lib/github-pages.rb
+++ b/lib/github-pages.rb
@@ -31,6 +31,7 @@ class GitHubPages
       "jekyll-sitemap"            => "0.9.0",
       "jekyll-feed"               => "0.3.1",
       "jekyll-gist"               => "1.4.0",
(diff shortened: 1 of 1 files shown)


parkr:
:shipit:

Jekyll-paginate will need to be in your 3.x PR, but perhaps it can be slipped in here, like jekyll-gist?

benbalter:
Looks like the latest release is 1.1.0? Meaning you're just asking that we explicitly version it?

Also, from the Jekyll Paginate description:

> RETIRED as of Jekyll 3: Pagination Generator for Jekyll

parkr:
It's not automatically included in Jekyll 3, so if it is to be available, it will have to be added as a plugin.

PR #187, opened 2016-01-12
Add Rubocop

benbalter:
Since I'm on a recent Rubocop kick (excluding a handful of tests that we're currently failing).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 .bundle
 CNAME
 test-site
+/test-site
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,135 @@
+# Ruby linting configuration.
+# See https://github.com/styleguide/ruby for the Ruby style guide
+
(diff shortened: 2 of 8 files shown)


bin/github-pages
```diff
@@ -1,54 +1,54 @@
 #!/usr/bin/env ruby
 
 require 'mercenary'
-require_relative "../lib/github-pages"
+require_relative '../lib/github-pages'
```
parkr: I thought it was double quotes everywhere. Let's get a rubocop file going for [GitHub's Ruby Styleguide](https://github.com/styleguide/ruby)

benbalter: TIL'd. Makes sense.

parkr:
@benbalter you might want to do a new pr - looks like there are some conflicts here

benbalter:
> looks like there are some conflicts here

Fixed. Assuming tests passed, I believe this should be good to go (I used your config)

parkr:
:shipit:

Issue #200, opened 2016-02-04
make it easy to configure local Jekyll to be Pages-like


afeld:
There were several changes in [the Jekyll 3.0 changeover](https://github.com/blog/2100-github-pages-now-faster-and-simpler-with-jekyll-3-0) that means that Jekyll needs a specific configuration to act the same locally as when the site is pushed to pages. A couple ideas about how to go about this:

* List the required configuration. As far as can tell, this would be

    ```yaml
    markdown: kramdown
    kramdown:
      input: GFM
    highlighter: rouge
    ```

* Offer a `github-pages serve` and `github-pages build`, which are a pass-through to the equivalent Jekyll commands, but with the `_config.yml` values overridden.

Thoughts?

benbalter:
> List the required configuration.

@afeld great minds and all that. We're brainstorming technical solutions, but in the mean time, see https://help.github.com/articles/using-jekyll-with-pages/#configuring-jekyll

parkr:
Yes! We will also need to include the gem whitelist so that folks running with `--safe` will still have their plugins required.

MattiSG:
> Offer a `github-pages serve` and `github-pages build`

:+1:

benbalter:
This is in progress over in https://github.com/github/pages-gem/issues/209.

Issue #208, opened 2016-02-16
Pagination doesn't work in safe mode


TWiStErRob:
Steps to reproduce:
 * `jekyll new .`
 * append to `config.yml`:
	```yml
	gems:
	  - jekyll-paginate

	paginate: 2
	paginate_path: "/page:num/"
	safe: false
	```
 * change `index.html`:
	```diff
	-{% for post in site.posts %}
	+{% for post in paginator.posts %} 

	+<nav id="pagination">
	+    {% if paginator.previous_page %}
	+      {% if paginator.previous_page == 1 %}
	+        <a href="{{ site.baseurl }}/">&laquo;</a>
	+      {% else %}
	+        <a href="{{ site.baseurl }}/page{{ paginator.previous_page }}/">&laquo;</a>
	+      {% endif %}
	+    {% endif %}
	+
	+    {% if paginator.next_page %} 
	+      <a href="{{ site.baseurl }}/page{{ paginator.next_page }}/">&raquo;</a>
	+    {% endif %} 
	+</nav>
	```
 * create some posts (I duplicated `welcome-to-jekyll` a few times)

Test the site and notice that pagination works nice.

Now, https://help.github.com/articles/using-jekyll-with-pages/#configuration-settings-you-cannot-change states that `safe: true` is mandatory and cannot be changed. But to have pagination I need to turn `safe: false`, otherwise it won't work: with safe mode enabled the generated index.html is empty, it doesn't show the posts.

How can I test the page locally with the allowed optional gems in safe mode to be closest to GitHub pages builds?

benbalter:
@parkr is possible with `safe: true` and no whitelist, all plugins are blacklisted?

TWiStErRob:
Ah, that actually makes sense :) Adding
```yaml
whitelist:
  - jekyll-paginate
```
allows me to test locally with safe mode on.

Would it be possible to incorporate this into the gem to automatically build in a github-like environment? Even at the cost of calling `github-gem serve` instead of `jekyll serve`.

benbalter:
> Would it be possible to incorporate this into the gem to automatically build in a github-like environment? Even at the cost of calling github-gem serve instead of jekyll serve.

Yep. We're working on just that over in #209.

TWiStErRob:
Awesome, keep up the good work!

parkr:
#209 has shipped! Use v62 or later.

PR #209, opened 2016-02-16
Version default configuration and configuration overrides

benbalter:
This is an ~~early experiment/~~attempt to use Jekyll hooks to set default configuration and configuration overrides locally.

For it to work, you must have the following in your Gemfile:

```ruby
gem 'github-pages', group: :jekyll_plugins
```

To do so, this splits the gem into two classes, Configuration, and Dependencies.

## Todo

- [x] Must bump HealthCheck, which is now a Module
- [x] Default Gems are not currently being included/required

Fixes https://github.com/github/pages-gem/issues/200


lib/github-pages/configuration.rb
```diff
@@ -0,0 +1,59 @@
...
+    }.freeze
+
+    def self.set(site)
+      user_config_overridden = Jekyll.configuration(OVERRIDES)
+      config = Jekyll::Utils.deep_merge_hashes(DEFAULTS, user_config_overridden)
+
+      site.instance_variable_set '@config', config
+      config.keys.each do |key|
+        site.send("#{key}=", config[key]) if site.respond_to?("#{key}=")
+      end
```
parkr: Could you just do `Jekyll::Utils.deep_merge_hashes(site.config, overrides)` then iterate through defaults and do `site.config[key] ||= default_value`?

benbalter:
@parkr gave it a bit of a rewrite, and tests are actually passing now. Could use your critical :eyes: once again. If there's a better way to implement this, feel free to commit directly to the branch.

lib/github-pages/configuration.rb
```diff
@@ -0,0 +1,118 @@
...
+    # * GH_ENV
+    OVERRIDES = {
+      "lsi"         => false,
+      "safe"        => true,
+      "plugins"     => SecureRandom.hex,
+      "whitelist"   => PLUGIN_WHITELIST,
+      "highlighter" => "rouge",
+      "kramdown"    => {
+        "template"          => '',
+        'math_engine'       => 'mathajx',
```
parkr: `mathjax`?

lib/github-pages/configuration.rb
```diff
@@ -0,0 +1,118 @@
...
+    # Set the site's configuration  Implemented as an `after_reset` hook.
+    def self.set(site)
+      config = Configuration.effective_config(site)
+
+      # Write the final config to the site object, noting that some values may
+      # have already been set as instancee variables when initialized
+      site.instance_variable_set '@config', config
+      config.keys.each do |key|
+        site.public_send("#{key}=", config[key]) if site.respond_to?("#{key}=")
+      end
```
parkr: This doesn't allow Jekyll to handle configs normally – `Jekyll.configuration` and `Jekyll::Site#initialize` do important things we'd have to replicate here. I would like to try something a bit less heavy-handed, maybe:

```ruby
def self.set(site)
  # Merge user config into defaults
  config = Jekyll::Utils.deep_merge_hashes(DEFAULTS, site.config)

  # Merge overwrites into user config
  Jekyll::Utils.deep_merge_hashes! config, OVERRIDES

  # Ensure we have those gems we want.
  config["gems"] = Array(config["gems"]) | DEFAULT_GEMS

  # Assign everything to the site
  Jekyll::Utils.deep_merge_hashes! site.config, config
end
```

Where does this fall short?

benbalter: What about [the instance variables set when the site is initialized](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/site.rb#L22-L25)?

parkr: You're right, those will have to be changed as well. But the above code works save those setters, right?

benbalter:
@parkr one last round of experimentation. I added a new `effective_config` method, the thinking being, is that rather than us implementing the same logic twice, we could use the method in production to produce the site's defaulting and overridden config, we just wouldn't run set, like we could locally (or is this the wrong approach?).

parkr:
Fixes #208, too.

lib/github-pages.rb
```diff
@@ -1,61 +1,11 @@
...
+end
 
-      # Gem versions we're curious about
-      "github-pages"  => VERSION.to_s,
-      "html-pipeline" => HTML::Pipeline::VERSION,
-      "sass"          => Sass.version[:number],
-      "safe_yaml"     => SafeYAML::VERSION
-    }
-  end
+Jekyll::Hooks.register :site, :after_reset do |site|
```
parkr: Why do this `after_reset`? `after_init` would be preferable, otherwise you're re-reading the configuration for every regeneration and causing a lot of GC pauses. Could slow things down quite a bit. Also doesn't take into account the side-effects of creating a site.

lib/github-pages/configuration.rb
```diff
@@ -0,0 +1,118 @@
...
+    #
+    # Note: this is a highly modified version of Jekyll#configuration
+    def self.effective_config(user_config, environment_overrides={})
+      # Jekyll defaults < GitHub Pages defaults
+      defaults = Jekyll::Utils.deep_merge_hashes(Jekyll::Configuration::DEFAULTS, DEFAULTS)
+
+      # Our overrides, optionally with environment-specific overrides passed at runtime
+      overrides = Jekyll::Configuration[OVERRIDES].stringify_keys
+      environment_overrides = Jekyll::Configuration[environment_overrides].stringify_keys
+      overrides = Jekyll::Utils.deep_merge_hashes(overrides, environment_overrides)
```
parkr: Please produce these overrides in a different method.

lib/github-pages/configuration.rb
```diff
@@ -0,0 +1,118 @@
...
+
+      # defaults < _config.yml < OVERRIDES
+      config = Jekyll::Configuration[defaults]
+      config = config.read_config_files(config.config_files(overrides))
+      config = Jekyll::Utils.deep_merge_hashes(config, overrides).stringify_keys
```
parkr: Don't forget to [`fix_common_issues`, `backwards_compatibilize`, and `add_default_collections`](https://github.com/jekyll/jekyll/blob/v3.0.3/lib/jekyll/configuration.rb#L189).

lib/github-pages.rb
benbalter: What do you mean by [`after_init`](https://github.com/jekyll/jekyll/search?utf8=%E2%9C%93&q=after_init)? I was going off the hooks on http://jekyllrb.com/docs/plugins/#hooks, hoping to get in to the config before plugins start getting required. Maybe we need a new, earlier hook?

lib/github-pages/configuration.rb
```diff
@@ -0,0 +1,118 @@
...
+
+      # Jekyll's native deep_merge_hashes doesn't merge arrays.
+      # Include default Gems, even if not requested, to avoid breaking pre 3.0 sites
+      config["gems"] = (config["gems"] | DEFAULT_PLUGINS).uniq
```
parkr: This requires every gem, even if it's not used. This is not how Pages works at present. It only requires a certain number of gems by default, leaving the vast majority as opt-in.

lib/github-pages/configuration.rb
```diff
@@ -0,0 +1,118 @@
...
+
+      config
+    end
+
+    # Set the site's configuration  Implemented as an `after_reset` hook.
+    def self.set(site)
+      environment_overrides = {}
+      LOCAL_PASS_THROUGH.each { |key| environment_overrides[key] = site.config[key] }
+      config = Configuration.effective_config(site.config, environment_overrides)
```
parkr: Perhaps `local_overrides`? I was expecting to see `ENV` somewhere in here.

lib/github-pages/configuration.rb
```diff
@@ -0,0 +1,118 @@
...
+
+      # Write the final config to the site object, noting that some values may
+      # have already been set as instancee variables when initialized
+      site.instance_variable_set '@config', config
```
parkr: :sob: this makes me sad. We should really do this differently in Jekyll, but there's never been a need for wholesale reconfiguration before. I wish it were simpler for us to do this.

spec/github-pages-configuration_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+      it "sets default gems" do
+        expect(config["gems"]).to include("jekyll-coffeescript")
+      end
+
+      it "lets the user specify additional gems" do
+        expect(config["gems"]).to include("jekyll-sitemap")
+      end
+
+      it "honors the user's config" do
```
parkr: This should also test at least one value in `LOCAL_OVERRIDES`

lib/github-pages.rb
parkr: ... `.khanify everything`

i despair. we cannot be re-reading all this stuff every time `reset` is called because it's called every time `process` is run. we need a new hook.

lib/github-pages/configuration.rb
benbalter: Wouldn't that have happened as part of `config.read_config_files` a few lines up when we read in the user's config?

lib/github-pages/configuration.rb
benbalter: The intention here was to build an array that concats the user's specified Gems + our own defaults. Is that not what's going on here?

lib/github-pages/configuration.rb
benbalter: Agreed, but this is the best I could do to Get It To Work (TM). What we really need, instead of a `hook`, is a `filter` or something similar, towards the end of [`read_config_files`](https://github.com/jekyll/jekyll/blob/v3.0.3/lib/jekyll/configuration.rb#L175), that allow us to modify the user's config as it's read, before the site object starts acting on it.

lib/github-pages.rb
benbalter: What if we put a flag on the site object to know if it's been overridden and bail if it has?

benbalter:
@parkr thanks for the :eyes:. Believe I implemented everything you suggested (save rewriting Jekyll :smile:).

benbalter:
@parkr I'll get the merge conflict resolved momentarily, but how are you feeling about this as it stands (we can also go back with a new hook in e.g., 3.2, to improve performance).

lib/github-pages/configuration.rb
parkr: So, to summarize, I'd see us doing:

```ruby
def self.set(site)
  # Merge user config into defaults
  config = Jekyll::Utils.deep_merge_hashes(DEFAULTS, site.config)

  # Merge overwrites into user config
  Jekyll::Utils.deep_merge_hashes! config, OVERRIDES

  # Ensure we have those gems we want.
  config["gems"] = Array(config["gems"]) | DEFAULT_GEMS

  # Assign everything to the site
  Jekyll::Utils.deep_merge_hashes! site.config, config

  # Update any instance variables
  site.config.each do |key, value|
    site.public_send("#{key}=", value) if site.respond_to?("#{key}=")
  end
end
```

parkr:
Running this with `jekyll-github-metadata` gives you an issue,  `Liquid Exception: undefined method `versions' for GitHubPages:Module in versions.json`, because we're [using the old version](https://github.com/jekyll/github-metadata/blob/381588afe6fdc61ed0b7b7b94d03d13e124e5952/lib/jekyll-github-metadata.rb#L70-L71).

@benbalter Could we add `GitHubPages.versions` as an alias to `GitHubPages::Dependencies.versions`, or should I update the plugin?

benbalter:
> Could we add GitHubPages.versions as an alias to GitHubPages::Dependencies.versions, or should I update the plugin?

Makes sense. :+1:

parkr:
@benbalter Which one?

benbalter:
> Which one?

Yes. :smile:

I'm fine with the alias, fewer breaking changes, fewer problems.

parkr:
@benbalter Hehe ok! Discovered a bug when testing, so thanks for that! https://github.com/jekyll/github-metadata/pull/42

But otherwise my method seems to work fine. What do you think about using it?

parkr:
@benbalter What do you think about this?

lib/github-pages/configuration.rb
```diff
@@ -0,0 +1,115 @@
...
+
+      # Ensure we have those gems we want.
+      config["gems"] = Array(config["gems"]) | DEFAULT_PLUGINS
+
+      config
+    end
+
+    # Set the site's configuration  Implemented as an `after_reset` hook.
+    def self.set(site)
+      return if site.config["_github_pages_processed"] == true
```
benbalter: What would happen if a malicious user put `_github_pages_processed: true` in their `_config.yml`? Should we either store it as an instance variable (is that safe?) or override the flag when we override the user's config? If the latter, should we use the existing `github` namespace?

benbalter:
Nice. I really like your approach. Made one comment on the flag. :soon:.

lib/github-pages/configuration.rb
parkr: Fixed in d2b8337.

benbalter:
Nice. A lot cleaner than my implementation. If you're happy, I say :shipit:.

PR #237, opened 2016-03-09
Add script/fmt & tell script/cibuild to use it

parkr:
Makes it easier to run `rubocop -a` to autocorrect offenses without having to worry about bundler.

:smile:


benbalter:
:sheep: it

PR #248, opened 2016-03-17
Add missing files to Gemspec

benbalter:


parkr:
@benbalter Wow, that's not good. What do you think about letting `git` handle this? It'll give us more flexibility to move forward.

```ruby
all_files       = `git ls-files -z`.split("\x0")
s.files         = all_files.grep(%r{^(bin|lib)/|^.rubocop.yml$})
s.executables   = all_files.grep(%r{^bin/}) { |f| File.basename(f) }
```

benbalter:
> What do you think about letting git handle this? It'll give us more flexibility to move forward.

Sounds good. Wasn't sure if we were going to be adding more files, and just wanted to get it working, since the dependencies weren't actually being distributed as part of the Gem, but feel free to PR that if you want. :smile:

PR #249, opened 2016-03-17
Collect files and executables in gemspec via git

parkr:
Follow-up to https://github.com/github/pages-gem/pull/248#issuecomment-197935302.

/cc @github/pages


benbalter:
:boat: it.

PR #251, opened 2016-03-17
Add dependency-configuration sanity checks

benbalter:
Now that we're versioning the config this PR adds a few minor sanity checks to ensure that:

* All default plugins are whitelisted
* All versioned plugins are whitelisted
* All whitelisted plugins are versioned


parkr:
Heck yes! :shipit:

PR #255, opened 2016-03-18
Update README to describe installing via Bundler as a Jekyll plugin

benbalter:
So that users get the local-config special sauce.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,17 +7,13 @@ A simple Ruby Gem to bootstrap dependencies for setting up and maintaining a loc
 
 ## Usage
 
-Run the following command:
+1. Add the following to your project's Gemfile:
 
-```
-gem install github-pages
-```
-
-Alternatively, you can add the following to your project's Gemfile:
-
-```ruby
-gem 'github-pages'
-```
+  ```ruby
+  gem 'github-pages', group: :jekyll_plugins
+  ```
+  
+2. Run `bundle install`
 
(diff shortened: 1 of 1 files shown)


budparr:
What does `group: :jekyll_plugins` do?

benbalter:
@budpar that tells Jekyll to load the gem as a plugin, somewhat similar to if you added it to `gems` in your config. In our case, that allows us to register a hook with Jekyll, regardless of your config, to inject our custom defaults and overrides.

parkr:
:shipit:

budparr:
Cool, thank you. :smile:

silvenon:
@benbalter with this we don't have to list Jekyll plugins under `gems` in `_config.yml`?

benbalter:
> with this we don't have to list Jekyll plugins under gems in _config.yml?

Correct-ish. You won't have to list [the default plugins](https://github.com/github/pages-gem/blob/master/lib/github-pages/configuration.rb#L5-L11).

dideler:
I added the gem to `:jekyll_plugins` group and bundled.

One of the default plugins is `jekyll-seo-tag`, so I removed that gem from my Jekyll config.

Now when I `jekyll serve` it fails because of the `{% seo %}` tag that has to be placed inside head.
```
jekyll 3.4.3 | Error:  Liquid syntax error (line 2): Unknown tag 'seo'
```
Am I misunderstanding that it would load the seo plugin?

parkr:
@dideler Yes, the way Liquid works is that the tags must be manually registered (which you do when you add `jekyll-seo-tag` to your Jekyll config), and the existence of a tag is checked when your document is parsed. No auto-loading occurs.

PR #258, opened 2016-03-19
Disable whitelist with DISABLE_WHITELIST flag

benbalter:
This will allow users to run non-whitelisted plugins locally, by passing the `DISABLE_WHITELIST` flag, which will whitelist any plugins included in the gem list.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,7 @@
 source "https://rubygems.org"
 
 gemspec
+
+gem "jekyll_test_plugin_malicious",
+  :github => "jekyll/jekyll-test-plugin-malicious",
+  :branch => "safe-flag"
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,3 @@
 source "https://rubygems.org"
 
 gemspec
-
-gem "jekyll_test_plugin_malicious", :github => "jekyll/jekyll-test-plugin-malicious"
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,4 @@ source "https://rubygems.org"
 
 gemspec
 
-gem "jekyll_test_plugin_malicious",
-  :github => "jekyll/jekyll-test-plugin-malicious",
-  :branch => "safe-flag"
+gem "jekyll_test_plugin_malicious", :github => "jekyll/jekyll-test-plugin-malicious"
--- a/lib/github-pages/configuration.rb
+++ b/lib/github-pages/configuration.rb
@@ -84,7 +84,7 @@ module GitHubPages
       end
 
       def disable_whitelist?
-        Jekyll.env == "development" && ENV["DISABLE_WHITELIST"] == "1"
+        Jekyll.env == "development" && !ENV["DISABLE_WHITELIST"].to_s.empty?
       end
 
       # Given a user's config, determines the effective configuration by building a user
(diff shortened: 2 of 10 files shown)


spec/github-pages-configuration_spec.rb
```diff
@@ -1,10 +1,18 @@
...
+      "source" => fixture_dir,
+      "quiet" => true,
+      "testing" => "123",
+      "destination" => tmp_dir }
+  end
   let(:configuration) { Jekyll.configuration(test_config) }
   let(:site)          { Jekyll::Site.new(configuration) }
   let(:effective_config) { described_class.effective_config(site.config) }
+  before { ENV.delete("DISABLE_WHITELIST") }
+  before { ENV["JEKYLL_ENV"] = "test" }
```
parkr: `before` is generally seen as bad to use, because it's not specific. `before(:all)` or `before(:each)` would be preferable.

lib/github-pages/configuration.rb
```diff
@@ -83,6 +83,10 @@ def processed(site)
         site.instance_variable_set :@_github_pages_processed, true
       end
 
+      def disable_whitelist?
+        Jekyll.env == "development" && ENV["DISABLE_WHITELIST"] == "1"
```
parkr: What if I write `DISABLE_WHITELIST=true`? That'd fail here. Could we be more flexible on its value?

Gemfile
```diff
@@ -1,3 +1,7 @@
 source "https://rubygems.org"
 
 gemspec
+
+gem "jekyll_test_plugin_malicious",
+  :github => "jekyll/jekyll-test-plugin-malicious",
+  :branch => "safe-flag"
```
parkr: [v0.2.0 released](https://github.com/jekyll/jekyll-test-plugin-malicious/releases/tag/v0.2.0).

parkr:
@benbalter still working on this?

benbalter:
@parkr pushed up updates for your comments

Gemfile
```diff
@@ -1,3 +1,5 @@
 source "https://rubygems.org"
 
 gemspec
+
+gem "jekyll_test_plugin_malicious", :github => "jekyll/jekyll-test-plugin-malicious"
```
parkr: Can this be removed?

spec/fixtures/_config.yml
```diff
@@ -2,4 +2,6 @@ some_key: some_value
 safe: false
 gems:
   - jekyll-sitemap
+  - jekyll_test_plugin_malicious
 quiet: false
+repository: "github/pages-gem"
```
parkr: Didn't we have a PR that grabbed this from the `git remote-url`?

Gemfile
benbalter: It doesn't appear to be in Ruby gems?

parkr: Huh? https://rubygems.org/gems/jekyll_test_plugin_malicious/versions/0.2.0

benbalter: Had 2.0, not 0.2. Fixed.

spec/fixtures/_config.yml
benbalter: https://github.com/jekyll/github-metadata/pull/45

benbalter:
@parkr 👀 good to you?

parkr:
@benbalter :shipit:

PR #276, opened 2016-04-19
Backwards-compatibilize user configuration

parkr:
Previous to this PR, you wouldn't be able to use `plugins`, `layouts` or any of the other deprecated options. They worked locally because Jekyll uses `Jekyll.configuration` to load the user's configuration, but once you tried on GitHub Pages, it would break because it's not using this same method.

/cc @github/pages


benbalter:
:shipit:

PR #316, opened 2016-08-23
Whitelist Jekyll Admin locally

benbalter:
This PR creates a new class of whitelisted plugins called "development plugins", meaning they can be run locally, without disabling the whitelist, but won't run in other environments.

The idea being that we want users to be able to run Jekyll Admin locally, by simply adding it to their Gemfile, but there's not a use case for ever needing it in production, since it's designed to be used locally (thus no need to whitelist generally).

/cc @mertkahyaoglu for FYI, and @parkr for 👓

### DIFF
--- a/lib/github-pages/configuration.rb
+++ b/lib/github-pages/configuration.rb
@@ -91,7 +91,7 @@ module GitHubPages
       end
 
       def disable_whitelist?
-        Jekyll.env == "development" && !ENV["DISABLE_WHITELIST"].to_s.empty?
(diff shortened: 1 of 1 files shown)


lib/github-pages/configuration.rb
```diff
@@ -89,6 +94,10 @@ def disable_whitelist?
         Jekyll.env == "development" && !ENV["DISABLE_WHITELIST"].to_s.empty?
```
parkr: Please use `development?` here now that the helper method exists. 😄

parkr:
LGTM other than the one comment above. 👍 🎉

PR #340, opened 2016-10-28
Use Ruby 2.3.1

parkr:
That's what we're runnin'! https://pages.github.com/versions/


DirtyF:
If I read the tests correctly, it needs a comment in the spec helper to freeze literals for them to pass:

`# frozen_string_literal: true`

parkr:
Thanks, Ben!

PR #352, opened 2016-11-08
Only exclude 'node_modules'. Do not exclude 'vendor'

parkr:
Many GitHub Pages users use `vendor/` to store third-party JavaScript, CSS, and other assets they incorporate into their sites. It is a widely-used standard for storing these kinds of assets.

A quick fix is to move all vendored assets data to the `assets/` directory, but that means the user is required to intervene to fix something that previously worked.

/cc @github/pages

### DIFF
--- a/lib/github-pages/configuration.rb
+++ b/lib/github-pages/configuration.rb
@@ -37,6 +37,7 @@ module GitHubPages
       "jailed"   => false,
       "gems"     => DEFAULT_PLUGINS,
       "future"   => true,
+      "exclude"  => %w{node_modules},
       "kramdown" => {
         "input"     => "GFM",
         "hard_wrap" => false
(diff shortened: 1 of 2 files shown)


mattr-:
Why not `vendor`?

parkr:
> Why not vendor?

Hey @mattr-! I updated the OP with more details.

mattr-:
>Hey @mattr-! I updated the OP with more details.

Cool. Thanks! :+1: from me - except for the failing build. :smiley:

afeld:
I worry that this would cause people to run into https://github.com/jekyll/jekyll/issues/5359 again...is there a way to avoid both problems?

benbalter:
1. Why make this change here, and not upstream?

2. Can we be more explicit and exclude `vendor/bundle`?

parkr:
Going to try to fix this upstream in a 3.3.1.

Issue #351, opened 2016-11-05
gh-pages version and jekyll version problem


DarkHollow:
I'm having a version problem when I do install or update. Instead of getting gh-pages 104 and jekyll 3.3.0, I'm getting 35/39 and 2.4.0

I do as follows

    $ jekyll new darkhollow.github.io

And get

    New jekyll site installed in /Users/roberto/Proyectos/darkhollow.github.io. 
    Running bundle install in /Users/roberto/Proyectos/darkhollow.github.io... 
    Fetching gem metadata from https://rubygems.org/...........
    Fetching version metadata from https://rubygems.org/..
    Fetching dependency metadata from https://rubygems.org/.
    Resolving dependencies...
    Using public_suffix 2.0.3
    Using colorator 1.1.0
    Using ffi 1.9.14
    Using forwardable-extended 2.6.0
    Using sass 3.4.22
    Using rb-fsevent 0.9.8
    Using kramdown 1.12.0
    Using liquid 3.0.6
    Using mercenary 0.3.6
    Using rouge 1.11.1
    Using safe_yaml 1.0.4
    Using minima 2.0.0
    Using bundler 1.13.6
    Using addressable 2.5.0
    Using rb-inotify 0.9.7
    Using pathutil 0.14.0
    Using jekyll-sass-converter 1.4.0
    Using listen 3.0.8
    Using jekyll-watch 1.5.0
    Using jekyll 3.3.0
    Using jekyll-feed 0.8.0
    Bundle complete! 3 Gemfile dependencies, 21 gems now installed.
    Use `bundle show [gemname]` to see where a bundled gem is installed.

And then

    $ cd darkhollow.github.io/
    $ vim Gemfile 

and I comment the line: `gem "jekyll", "3.3.0"`
and I uncomment the line: `gem "github-pages", group: :jekyll_plugins`

Finally I do

    $ bundle update

and I get this

    Fetching gem metadata from https://rubygems.org/..................
    Fetching version metadata from https://rubygems.org/...
    Fetching dependency metadata from https://rubygems.org/..
    Resolving dependencies...
    Using RedCloth 4.2.9
    Using concurrent-ruby 1.0.2
    Using i18n 0.7.0
    Using minitest 5.9.1
    Using thread_safe 0.3.5
    Using addressable 2.4.0 (was 2.5.0)
    Using blankslate 2.1.2.4
    Using fast-stemmer 1.0.2
    Using coffee-script-source 1.10.0
    Using execjs 2.7.0
[...]

PR #353, opened 2016-11-08
warn about bundler issue in README

jldec:
Trying to help users avoid #351 
(caused by https://github.com/bundler/bundler/issues/5154)

ready for 👀

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,10 +7,11 @@ A simple Ruby Gem to bootstrap dependencies for setting up and maintaining a loc
 
 ## Usage
 
-1. Add the following to your project's Gemfile:
+1. Add the following to your project's Gemfile:  
+_Note: Forcing the version with `'>=104'` is a temporary workaround to avoid #351._
 
   ```ruby
-  gem 'github-pages', group: :jekyll_plugins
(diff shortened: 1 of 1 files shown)


jldec:
Remember to remove the warning and the version number from the README when Bundler releases a fix.

benbalter:
@parkr @jldec I think it's a mistake to tell users to lock the gem at version `104` as many users will fix the problem once and move on, forgetting about the flag. Can we update the instructions to be `>= 104` and achieve the same result?

jimkohl:
Ben
I first attempted as you suggest but that didn't work.  Using the specific version only worked.  

While I generally share your concern in this case many of us are well aware of this issue and awaiting the resolution to resolve correctly.  Spent too much time helping others Monday get around this. 
Tx for all you do here!
Jim

parkr:
@benbalter `>= 104` and `104` have the same effect: install v104. Sadly, `>= 104` doesn't work presently. That said, it's a better place to be once bundler *is* fixed.

benbalter:
What if we add `jekyll ~> 3.3` to force the higher resolution?

PR #384, opened 2017-01-11
Lock nokogiri to 1.6 because 1.7 dropped support for Ruby 2.0

parkr:
Ruby 2.0 is the version of Ruby that is installed by default on Mac OS X. Many of our users utilize this and we should do our best to keep compatibility.


parkr:
Bitten by https://github.com/sickill/rainbow/pull/46 again. Womp.

PR #388, opened 2017-01-14
Bump jekyll-theme-primer to v0.1.7

hubot:
**Release**: https://github.com/pages-themes/jekyll-theme-primer/releases/tag/v0.1.7
**Diff**: https://github.com/pages-themes/jekyll-theme-primer/compare/v0.1.5...v0.1.7

PR #389, opened 2017-01-14
Bump versions

DirtyF:
☝️


DirtyF:
CI is not passing because of Rubocop, fixed in #388

parkr:
@DirtyF I knew that I had added that comment so I was wondering where it went! Looks like the script we use to auto-bump the version and release, `script/bump`, needed to be updated as well: https://github.com/github/pages-gem/pull/390

jldec:
Merged via #392 - thanks again.

PR #396, opened 2017-01-21
Workaround for bundler no more needed

DirtyF:
Everything works fine with bundler 1.14 now :tada:


DirtyF:
closes ##396

parkr:
I think it would be wise to include a note here which asks folks to upgrade to Bundler 1.14!

DirtyF:
@parkr better?

PR #398, opened 2017-01-24
Bump kramdown to v1.13.2

hubot:
**Release**: http://github.com/gettalong/kramdown/releases/tag/REL_1_13_2
**Diff**: http://github.com/gettalong/kramdown/compare/REL_1_11_1...REL_1_13_2
**Changelog**: https://github.com/gettalong/kramdown/blob/master/doc/news/release_1_12_0.page
**Changelog**: https://github.com/gettalong/kramdown/blob/master/doc/news/release_1_13_0.page

PR #400, opened 2017-01-31
Bump kramdown to v1.13.2, and also explicitly set gfm_quirks option to paragraph_end

neon-ninja:
Enabling gfm_quirks by default as mentioned by @benbalter in https://github.com/github/pages-gem/pull/398

### DIFF
--- a/lib/github-pages/configuration.rb
+++ b/lib/github-pages/configuration.rb
@@ -19,6 +19,7 @@ module GitHubPages
       "kramdown" => {
         "input"     => "GFM",
         "hard_wrap" => false,
+        "gfm_quirks" => "paragraph_end",
       },
     }.freeze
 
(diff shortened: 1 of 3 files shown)


lib/github-pages/configuration.rb
```diff
@@ -19,6 +19,7 @@ class Configuration
       "kramdown" => {
         "input"     => "GFM",
         "hard_wrap" => false,
+        "gfm_quirks" => "paragraph_end",
```
parkr: Should this simply be an enforced override? Under what circumstances would a user want to turn this feature off?

benbalter: I can't think of any, other than perhaps wanting to change other flags down the line (which would be in that same string).

PR #408, opened 2017-02-08
Add an automatic version check.

parkr:
Ensuring that you're up to date with the latest github-pages version means you'll be accurately building your site locally to match the eventual Pages production deploy.

@benbalter, what do you think about this?

### DIFF
--- a/lib/github-pages/version_check.rb
+++ b/lib/github-pages/version_check.rb
@@ -1,0 +1,58 @@
+# frozen_string_literal: true
+require "typhoeus"
+
+module GitHubPages
+  # VersionCheck ensures that the local version is up to date with the
+  # latest version deployed to GitHub Pages.
+  module VersionCheck
+    PRODUCTION_VERSIONS_URL = "https://pages.github.com/versions.json"
+
+    def run!
+      @already_run ||=
+        begin
+          return true if ENV["PAGES_ENV"].to_s.casecmp("enterprise").zero?
+
+          if GitHubPages::VERSION < production_ghp_version
+            Jekyll.logger.error "GitHub Pages:", "You are using an outdated version of github-pages."
+            Jekyll.logger.error "", "Upgrade to version '#{production_ghp_version}' to ensure your local builds"
+            Jekyll.logger.error "", "will accurately reflect what is deployed when you push to GitHub."
+            return false
+          end
+
+          true
+        end
+    end
+
+    def production_versions_url
+      @production_versions_url ||= PRODUCTION_VERSIONS_URL
+    end
+
+    def production_ghp_version
+      @production_ghp_version ||= production_versions["github-pages"].to_i
+    end
+
+    def production_versions
+      @production_versions ||=
+        begin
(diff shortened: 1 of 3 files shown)


benbalter:
I like it, but first and foremost, this can't ever be a blocker. If I don't have internet, GitHub Pages is down, the JSON is invalid, I have the wrong version, etc. it should silently fail. I believe we recently used a DNS check to quickly tell if the user had internet access. Perhaps we should do that here as well, to fail as quickly as possible (rather than adding a 30 second timeout).

lib/github-pages/version_check.rb
```diff
@@ -0,0 +1,58 @@
...
+    def run!
+      @already_run ||=
+        begin
+          return true if ENV["PAGES_ENV"].to_s.casecmp("enterprise").zero?
+
+          if GitHubPages::VERSION < production_ghp_version
+            Jekyll.logger.error "GitHub Pages:", "You are using an outdated version of github-pages."
+            Jekyll.logger.error "", "Upgrade to version '#{production_ghp_version}' to ensure your local builds"
+            Jekyll.logger.error "", "will accurately reflect what is deployed when you push to GitHub."
+            return false
```
benbalter: We should explain *how* to update and ideally link to the help docs.

lib/github-pages/version_check.rb
```diff
@@ -0,0 +1,58 @@
...
+          JSON.parse(response.response_body)
+        end
+    end
+
+    extend self
+
+    private
+
+    # rubocop:disable Lint/HandleExceptions
+    def http_get_with_retries(url)
```
benbalter: Why do we need to retry 5 times?

parkr: Using `open-uri`, I have experienced about half of the time pulling in this file and getting nothing back with some kind of server error. I figured we should retry just in case the user is on an unreliable connection such as coffeeshop wifi or airport wifi.

lib/github-pages/version_check.rb
parkr: Agreed. This link? https://help.github.com/articles/setting-up-your-github-pages-site-locally-with-jekyll/#keeping-your-site-up-to-date-with-the-github-pages-gem

parkr:
> this can't ever be a blocker

Agreed – a reliable internet-connectivity check is required. A DNS check *should* work, though a TCP connection would be an even better bet of ensuring our connection is capable of this kind of transaction.

> rather than adding a 30 second timeout

2 second timeout? This runs once right at the beginning.

kvz:
Stumbled upon this. In case it helps: I've seen in the Node.js ecosystem the way is to check during/after the main job is done, saving new version information to disk. This can fail all it wants. But if the next execution does find such a file on disk with a higher version number, it is proposed to the user.

adamvoss:
> 2 second timeout? This runs once right at the beginning.

If I am doing something where _watch_ does not work and I need to restart Jekyll to rebuild, 2 seconds is an awfully long time to need to wait on every start.

PR #423, opened 2017-03-09
Dockerization of GitHub Pages (WIP)

johndmulhausen:
Creating this PR as a starting point for making GitHub Pages run in a Docker container. 

Try this out by cloning my fork locally, `cd`ing into the repo and running:

```
docker build -t myghpages .
```

Then, because this isn't running yet, you'll want to start a bash prompt from within the container:

```
docker run -ti --entrypoint=/bin/bash myghpages
```

From here you can see what has happened up to the point of the `CMD` line (which is overridden by `entrypoint`). There's a built `github-pages-127.gem` in your directory, but it doesn't seem to install, probably because I'm a newb and am missing something obvious. From within that prompt try running:

```
bundle install --gemfile=github-pages-127.gem
```

...And you'll see what I mean:

```
[!] There was an error parsing `github-pages-127.gem`: Undefined local variable or method `metadata' for Gemfile. Bundler cannot continue.

 #  from /usr/src/app/github-pages-127.gem:1
 #  -------------------------------------------
 >  metadata.gz0000444000000000000000000000247413060347710013446 0ustar00wheelwheel00000000000000???X?ao?6???)X??͠?d?t?i?t?I?????&?'????6?E????1?k?kO?]?XM[??1??/?<~l?Q?_?v?=??K?&{u???ekF?aR?Ԑ?%3U??R??@iܔ=????? 
                              ?V????Y?pb?R?Yh?	Hk*?tD?~lW_?o?m?T?w?6??(??B?PP&?ax???
 #  
 #  -------------------------------------------
```

cc @thajeztah, @mstanleyjones, @parkr 

Just a start; the people on the CC line are definitely capable of creating this, and I appreciate any advice from you guys. I can also advise GitHub on how to get an official image for GitHub Pages up on the Docker Store, and get it auto-building so pushes to this repo update the official image. But first, let's get it working. :)

The "Dockerization" of GitHub Pages would be tremendous; currently there is no official image and people are using janky things from the community (like `starefossen/github-pages`) which don't pass [Docker Security Scanner's security checks](https://docs.docker.com/docker-cloud/builds/image-scan/).

> Note: Once this is building properly, the bits in `CMD` should be moved up to `RUN` and `CMD` should probably get a pointer to `jekyll`as the real entrypoint

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM ruby:2.3.3-alpine
+
+RUN mkdir /usr/src/app \
+  && apk update \
+  && apk upgrade \
+  && apk add --no-cache bash git make
+
+COPY . /usr/src/app
+
+WORKDIR /usr/src/app
+
+RUN gem build github-pages.gemspec
+
+CMD bundle install --gemfile=github-pages-127.gem


parkr:
Have you tried the official Jekyll images? One specifically bundles this gem, `jekyll/jekyll:pages`. https://github.com/jekyll/docker/wiki

johndmulhausen:
@parkr There are a few advantages this approach would have with `jekyll:pages`

- Versioning. If GitHub Pages had official Docker images, then each version could have a tagged image. You could run something like `FROM github/pages:v112` instead of being pinned to latest via `jekyll/jekyll:pages` which could break your site someday.
- Docker workflow for dev/test of `github-pages` becomes possible.
- More WYSIWYG: https://github.com/jekyll/docker/blob/master/opts.yml  <-- This actually creates somewhat unclear of a final environment, with many gem collisions happening. Both `pages` and `all` include many of the same gems, which can lead to gnarly things, especially when trying to run vanilla `jekyll serve` ([as the Dockerfile specifies](https://github.com/jekyll/docker/blob/master/repos/jekyll/Dockerfile#L39)).
- More well-maintained and no possibility of falling out of sync. It's... pretty quiet over at jekyll/docker. Most of the contents haven't been updated in a year. With this and a bit of push-button CI config, we can make sure every update pushes a new Docker image.

vidbina:
Hey @johndmulhausen, excuse me for just barging in here but I probably don't understand your goal well enough, so help me out. Personally, I like the idea of a test and development container that one could mostly use it to run whatever is already in the script directory :wink:, but with the entire `bundle install --gemfile` rule I'm getting the feeling that you are aiming for a usage container where you potentially could fire up `jekyll serve` to examine the output of jekyll in your browser but I'm hopelessly mistaken.

So please clarify whether you intend this to be a usage container (for firing up `jekyll serve` eventually) or a test/development container. I would say the `bundle install --gemfile` command is needed for neither of the two use-cases, and here's why I would think that:

If you just want to run a Jekyll site with an environment that closely matches your gh-pages environment, I would suggest considering running `bundle install` inside the gem directory and then subsequently changing to your jekyll site directory and just running some variation of `jekyll serve` there.
[asciicast](https://asciinema.org/a/ITl4nc124BgBP2PgoB6myW9o0)

The [bundle install](http://bundler.io/bundle_install.html) states
> Gems will be installed to your default system location for gems. If your system gems are stored in a root-owned location (such as in Mac OSX), bundle will ask for your root password to install them there.

which means that switching directories inside the container will do little in terms of changing which gems are available after the directory change. We're not using any tooling inside the container that changes the ruby and gem context as we change directories (i.e.: we're not honoring `ruby-version`, `ruby-gemset` or any of such files with a similar purpose).

Since [Github Pages dictates](https://pages.github.com/versions/) the environment we run on, I conveniently ignore the Gemfile or Gemfile.lock that may be part of the jekyll site repository. It's only there to allow the user to locally run the jekyll server or attempt a build so I figured it's reasonable to consider the pages-gem our single source of truth and ignore everything else.

So how about this...
 1. strip the container just to have the bare minimum in place for running all the scripts within the script directory
 2. copy the pages-gem directory into the container
 3. `bundle install` inside the pages-gem directory

Now the user can basically:
 1. fire up an instance with the pages-gem directory mounted into it (mounted such that changes to the fs are visible within the container without a restart), change into the pages-gem directory and run the scripts for test/development purposes or
 1. fire up an instance with a jekyll site directory mounted into it, change into the site directory and run `jekyll serve` to watch the fireworks 🎆 

But then again, as @parkr mentioned, Jekyll already provides the `pages` tag and I'm not so sure about the broken site argument, since an update to Github Pages is not generally optional. Any request for a github pages site build will make subjected the jekyll site to the manifest of dependencies used by Github Pages at the moment, so I guess people are better of using the lastest version `jekyll:pages` anyways, but even here my understanding is very likely limited ;). I understand the other argument and agree that our setup could be a lot less verbose and contain fewer surprises to the end user.

vidbina:
Added a [PR](https://github.com/johndmulhausen/pages-gem/pull/1) that demonstrates some of my proposed changes :wink:.

johndmulhausen:
Closing my PR in favor of @vidbina

PR #428, opened 2017-03-28
Bump outdated jekyll-* dependencies

benbalter:
### Jekyll Feed

* fix <entry> template for posts with post.lang defined 

### Jekyll Relative links

* Support for relative static files
* Save from 'invalid byte sequence' errors

### Jekyll README index

* Support for READMEs in subfolders
* README detection now respects global Markdown extension setting

### DIFF
--- a/lib/github-pages/version.rb
+++ b/lib/github-pages/version.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module GitHubPages
-  VERSION = 130
+  VERSION = 131
 end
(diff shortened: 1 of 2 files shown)


parkr:
We don't need to lock the source for coffeescript anymore near as I can tell. Worth skipping this version of jekyll-coffeescript and going back to it being unlocked?

lib/github-pages/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module GitHubPages
-  VERSION = 130
+  VERSION = 131
```
parkr: I'd rather see this happen automatically!

benbalter: > I'd rather see this happen automatically!

How?!?!

benbalter: Ohhhhhh

PR #453, opened 2017-07-17
Bump primer to 0.3.0

benbalter:
* Adds a default `H1` with `site.title` and a link to the site's index, if set
* Adds margin-top and margin-bottom to the body

### DIFF
--- a/lib/github-pages/plugins.rb
+++ b/lib/github-pages/plugins.rb
@@ -46,7 +46,7 @@ module GitHubPages
     THEMES = {
       "minima"                     => "2.1.1",
       "jekyll-swiss"               => "0.4.0",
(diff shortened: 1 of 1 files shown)


DirtyF:
@benbalter why isn't this theme publicly listed on https://github.com/pages-themes/ ?

benbalter:
> why isn't this theme publicly listed on https://github.com/pages-themes/ ?

We opened source the themes that used to be part of the theme chooser, but didn't open source Primer. The thinking at the time was that it'd be a disproportionate maintenance burden, since it's just (already open source) Primer Sass vendored + the simple layout. Not to mentioned, we didn't expect many users to want to customize the intentionally minimal template. If there's desire now, we can reevaluate that decision.

PR #456, opened 2017-07-17
Default sass to compressed in production

benbalter:
This PR sets the sass style to "compressed" in production by default (which can be user-overridden).

To implement this, I added a `PRODUCTION_DEFAULTS` hash, which contains the production-only defaults.

I also had to move `MERGED_DEFAULTS` from a constant to a method so that it could be dynamic (based on the jekyll environment).


spec/github-pages/configuration_spec.rb
```diff
@@ -159,6 +167,11 @@
           expect(described_class.disable_whitelist?).to eql(false)
         end
       end
+
+      it "compresses sass" do
+        puts ENV["JEKYLL_ENV"].inspect
```
parkr: ?

benbalter: Removed via 334f384.

lib/github-pages/configuration.rb
```diff
@@ -82,6 +83,11 @@ def development?
         Jekyll.env == "development"
       end
 
+      def merged_defaults
```
parkr: `defaults_for_env`? `env_default_configuration`? It'd be good to know that this method is sensitive to the environment.

PR #461, opened 2017-08-08
Dockerfile (with some convenience Makerules) for development, testing and serving

vidbina:
@johndmulhausen started this with github/pages-gem#423 which led to a subsequent PR into his set of changes johndmulhausen/pages-gem#1. Since his `master` branch is lagging, the diff contains a lot of noise (commits that are unrelated to the Dockerization things) as [pointed out](https://github.com/johndmulhausen/pages-gem/pull/1#issuecomment-320725726) by @benbalter. This PR minimizes the diff to something more suited for scrutiny.

# Changes

 - Cleaned up the Dockerfile to the bare minimum for using it as a test/dev environment and even using is as a server to view any jekyll-site in a github-pages-like environment
 - Set up a bundle override for the github-pages gem to point to the local directory
 - Added some make rules to simplify usage

# Usage

 - Use `make image` to build the github-pages image
 - Use `make shell` to produce a bash shell
   - run the scripts inside that shell by simply running
     - `./script/bootstrap`, usually run this first to ensure a fresh
       `bundle install` (note that with the `make shell` rule mounting the pages-gem directory into the container's `/gh-pages` path (the bundler override for the github-pages gem), one may test changes made to the directory on the host, rather than being limited to having to rebuild the image on every file change :wink:)
     - `./script/cibuild`
     - `./script/test-site`
     - `./script/fmt`
 - Run the server by (alternative approaches in order of preference):
   - prepending an assignment to `SITE` to the `make server` rule, e.g: `SITE=../path-to-jekyll-site make server`
   - exporting `SITE` first and then calling `make server`, e.g.: `export SITE=../path-to-jekyll-site; make server`. After the `export`, one may trigger the server by simply running `make server`, but the `SITE` assignment becomes less explicit. When you change to another terminal that perhaps already has the variable `SITE` defined, perhaps even for other purposes, one will basically be altering the terminal's state.

## Demo

[asciicast](https://asciinema.org/a/132541)


johndmulhausen:
This is great; I'll close my pr :)

benbalter:
@vidbina this is great. I'd love to get this merged. Would you be able to add documentation? If so, I believe this should be big improvement.

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+RUN apt-get update \
+  && apt-get install -y \
+    git \
+    make
+
+COPY . /gh-pages
+WORKDIR /gh-pages
+
+RUN bundle config local.github-pages /gh-pages
+RUN bundle install
```
benbalter: We may need to modify the Gemfile to add the Gem to the `jekyll_plugins` group so that it can hook earlier into the bootstrap process and set defaults, e.g., `gem "github-pages", group: :jekyll_plugins`

vidbina: I'm not sure I totally understand what you mean with this, please elaborate and I'll try to get on it asap :wink:.

My GitHub Pages projects do contain the `gem "github-pages", group: :jekyll_plugins` line, but then I believe I should then probably fire up "bundle exec jekyll server ..." instead of just  "jekyll serve ...".

Is there any logic performed on the `jekyll_plugins` group? Is it not only there for organization?

vidbina:
@johndmulhausen and @benbalter can you check if the documentation I added makes any sense at all?!? Thanks in advance 🙇

Dockerfile
benbalter: > Is there any logic performed on the jekyll_plugins group? Is it not only there for organization?

The Jekyll Plugins group allows the GitHub Pages Gem to get bootstrapped earlier in Jekyll's process and set configuration defaults and constraints to match GitHub Pages. I'm struggling to load a gemspec into a gem group, so we may need to create a new / modify the Gemfile for docker.

wolfgang42:
This image looks very exciting for allowing local testing of GitHub Pages sites. I've been testing it on the [GitHub Pages examples](https://github.com/showcases/github-pages-examples) and I've noticed some examples where it produces errors. All of these were produced by cloning the site and then running `SITE=... make server`.

* https://github.com/jekyll/jekyll
    ```
      Liquid Exception: invalid byte sequence in US-ASCII in _layouts/redirect.html
    jekyll 3.5.2 | Error:  invalid byte sequence in US-ASCII
    ```
* https://github.com/github/government.github.com
    ```
    jekyll 3.5.2 | Error:  Could not find a JavaScript runtime. See
    https://github.com/rails/execjs for a list of available runtimes.
    ```
* https://github.com/facebook/react/
    ```
      Dependency Error: Yikes! It looks like you don't have redcarpet or one of its 
    dependencies installed. In order to use Jekyll as currently configured, you'll
    need to install this gem. The full error message from Ruby is: 'cannot load such
    file -- redcarpet' If you run into trouble, you can find helpful resources at 
    https://jekyllrb.com/help/!
      Conversion error: Jekyll::Converters::Markdown encountered an error while
    converting '_posts/2013-06-02-jsfiddle-integration.md':
                    redcarpet
             ERROR: YOUR SITE COULD NOT BE BUILT:
                    ------------------------------------
                    redcarpet
    ```

I'm not sure if this is what's causing the above errors, but it looks like the way Jekyll has been set up doesn't match how GitHub Pages does rendering: Pages runs Jekyll in 'safe mode' with a specific whitelist of plugins, so sites which work in this container might not work on Pages if they're using custom plugins. I've found the settings for all of this in [`configuration.rb`](https://github.com/github/pages-gem/blob/master/lib/github-pages/configuration.rb) but I'm not sure how to hook that into Jekyll.

vidbina:
Hey @wolfgang42, thanks for this. I'll take some time soon enough to look into this issue. I've been flooded with some other chores meanwhile so I'll get back at you as soon as I know more :wink:. Thanks for listing the specific examples and the errors they produce.

Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM ruby:2.4
+
+RUN apt-get update \
+  && apt-get install -y \
+    git \
+    make
+
+COPY . /gh-pages
```
parkr: Due to how `Jekyll.sanitized_path` works, let's put this in a subdirectory 2 levels deep, e.g. `/gh/pages` or `/src/gh-pages`. Otherwise, if you have a file called `gh-pages.html`, you'll get an error. It's confusing so let's tackle that before merging.

Makefile
```diff
@@ -0,0 +1,30 @@
...
+		-p 4000:4000 \
+		-v ${PWD}:/gh-pages \
+		${TAG} \
+		/bin/bash
+
+# Spawn a server. Specify the path to the SITE directory by
+# exposing it using `expose SITE="../path-to-jekyll-site"` prior to calling or
+# by prepending it to the make rule e.g.: `SITE=../path-to-site make server`
+server:
+	ls "${SITE}" >/dev/null && \
```
parkr: Can an error message be printed here if the SITE variable doesn't exist?

Makefile
```diff
@@ -0,0 +1,30 @@
...
+	${DOCKER} run --rm -it \
+		-p 4000:4000 \
+		-v ${PWD}:/gh-pages \
+		-v `realpath ${SITE}`:/site \
+		-w /site \
```
parkr: Can we use `/src/site` (a subdirectory 2 levels deep) for reasons stated above?

README.md
```diff
@@ -19,6 +23,29 @@ A simple Ruby Gem to bootstrap dependencies for setting up and maintaining a loc
...
+2. Start an instance of the server by running either:
+  - `SITE=PATH_TO_YOUR_PROJECT make server` from the root of the `gh-pages` repository (where the Makefile resides) or
+  - ``SITE=PATH_TO_YOUR_PROJECT docker run --rm -p 4000:4000 -v `realpath ${SITE}`:/site gh-pages`` from any directory or
+  - `github-pages $PATH_TO_YOUR_PROJECT` from any directory when [func.sh](func.sh) has been sourced into your terminal session or
+  - `github-pages` from the directory of the Jekyll site to be previewed when [func.sh](func.sh) has been sourced into your terminal session.
+
+**Note:** the `github-pages` function may be enabled by sourcing func.sh. This can be done by appending
+
+  ```bash
+  source $PATH_TO_THIS_DIRECTORY/func.sh
```
parkr: This is great! Can it go in a `contrib/` subdirectory instead of the root of the project? `contrib/docker` maybe?

Dockerfile
vidbina: cbe4cdf54238210673752bdecf44ebc1e0c08d00

Makefile
vidbina: Will an `echo` suffice? 03709aa56f1062b7566ca604258af285ee46f2ef

Makefile
vidbina: Done :wink: 9d8fe1886fccaf4355df158dac74707eb87cec37

README.md
vidbina: Done :wink: f4a8d87bb0d74185b48d69d889e6940a683de132

vidbina:
@parkr I kind of took forever to get around to it, apologies. It's only now when I'm chilling on an island :palm_tree: that I could site for a few minutes and just work through some of your points. Still intent to address the issues addressed by @benbalter and @wolfgang42 but it could take a little while. Just bear with me and rest assured that I've heard and acknowledged the concerns loud and clear :ear:

vidbina:
@wolfgang42 just called `github-pages` inside the [jekyll](https://github.com/jekyll/jekyll) directory resulting to the following:

```
            Source: /src/site
       Destination: /src/site/_site
 Incremental build: disabled. Enable with --incremental
      Generating... 
             Error: could not read file /src/site/lib/site_template/_posts/0000-00-00-welcome-to-jekyll.markdown.erb: Invalid date '<%= Time.now.strftime('%Y-%m-%d %H:%M:%S %z') %>': Document 'lib/site_template/_posts/0000-00-00-welcome-to-jekyll.markdown.erb' does not have a valid date in the YAML front matter.
             ERROR: YOUR SITE COULD NOT BE BUILT:
                    ------------------------------------
                    Invalid date '<%= Time.now.strftime('%Y-%m-%d %H:%M:%S %z') %>': Document 'lib/site_template/_posts/0000-00-00-welcome-to-jekyll.markdown.erb' does not have a valid date in the YAML front matter.
```
Not sure why I doesn't evaluate the the erb template. Will look into it.

Running `github-pages` inside the [government.github.com](https://github.com/github/government.github.com) directory resulted to:

```
Configuration file: /src/site/_config.yml
            Source: /src/site
       Destination: /src/site/_site
 Incremental build: disabled. Enable with --incremental
      Generating... 
  Liquid Exception: invalid byte sequence in US-ASCII in _layouts/redirect.html
jekyll 3.6.2 | Error:  invalid byte sequence in US-ASCII
```

for which I initially figured that it could be related to the [old](https://github.com/jekyll/jekyll/issues/960#issuecomment-21311957) `LC_ALL` var issue, but I have set `LANG`, `LANGUAGE` and `LC_ALL` to no avail.

For [facebook/react](https://github.com/facebook/react/) I haven't found a jekyll site. Perhaps the link you referenced doesn't point to the static page for React?

Will have to do a bit of digging tomorrow to figure out what's up but perhaps @benbalter or @parkr have some insights that could nudge me in the right direction.

wolfgang42:
@vidbina It looks like the React Jekyll site was deleted last month: https://github.com/facebook/react/pull/10941,  https://github.com/facebook/react/commit/887ccf2534ab010bc645984ccff390a3e9f2c671

vidbina:
Hey @wolfgang42. The Jekyll site builds as long as you use the `docs` subdir as the site directory

`github-pages /path/to/jekyll-repository/docs` or `SITE=/path/to/jekyll-repository/docs make server` will do just fine.

The invalid byte sequence issue was basically resolved by configuring the locale of the container through the setting of the `LANG`, `LC_ALL` and `LANGUAGE` variables, modification of the `/etc/locale.gen` file and running of the `locale-gen` command. I probably overdid it but was too lazy to figure out which ones of these operations constitute the bare minimum to get the setup working.

As for the execjs issue, that was resolved by `apt-get install nodejs` as recommended in the [troubleshooting page](https://jekyllrb.com/docs/troubleshooting/#could-not-find-a-javascript-runtime-execjsruntimeunavailable) and [jekyll#2327](https://github.com/jekyll/jekyll/issues/2327).

Let me know if that solves the issues :wink:.

Dockerfile
liskin: Have you guys figures this out in the end? It seems to me that jekyll in the container ignores github-pages entirely and just uses its own default configuration, making this whole `make server` and `contrib/func.sh` thing rather useless. :-/

PR #473, opened 2017-08-24
Recommandations before suggesting a plugin

DirtyF:
GItHub Pages can't afford to add low-quality plugins to millions of users. Hopefully it's more clearly stated now.

Ideally gems should pass a security check with something similar to e.g. https://hakiri.io/


benbalter:
Nice. Thanks for this, @DirtyF!

PR #477, opened 2017-08-31
Add basic integration tests

benbalter:
This PR adds some basic integration tests to ensure that:

1. Every bundled plugin actually runs
2. The output is roughly as expected

It's not intended to be a fully comprehensive test of each plugin's functionality, as each plugin should theoretically do that up stream, but these basic tests should give us an early warning if a dependency bump breaks a plugin.

The tests live in `spec/integration_spec.rb` with each dependency in its own context. Each dependency also has its own fixture to keep tests distinct and organized. The spec builds the fixture site and then verifies the output of each file matches expectations. There are some helpers baked in if tests follow this pattern, but strictly speaking, it's not necessary.

There's also a simple regex-based test to ensure that newly added dependency get integration tests.

Last, I'm not 100% sure why the default theme isn't working (perhaps we hook too late), but this is a start.


spec/spec_helper.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require File.expand_path("../../lib/github-pages.rb", __FILE__)
```
DirtyF: `require File.expand_path("../lib/github-pages.rb", __dir__)`

spec/spec_helper.rb
```diff
@@ -14,9 +15,13 @@
 end
 
 def fixture_dir
-  File.expand_path "./fixtures", File.dirname(__FILE__)
+  File.expand_path "./fixtures", __dir__
```
DirtyF: you don't need to prefix with `./` here:
`  File.expand_path "fixtures", __dir__`

spec/github-pages/integration_spec.rb
```diff
@@ -0,0 +1,215 @@
...
+  def env
+    { "BUNDLE_GEMFILE" => "#{source}/Gemfile" }
+  end
+
+  def run_or_raise(env, *cmd)
+    stdout_and_stderr_str, status = Open3.capture2e env, *cmd
+    raise StandardError, stdout_and_stderr_str if status.exitstatus != 0
+  end
+
+  let(:file) { "#{self.class.description}.html" }
```
parkr: ?

spec/github-pages/integration_spec.rb
```diff
@@ -0,0 +1,215 @@
...
+  let(:path) { destination_file file }
+  let(:contents) { File.read path }
+
+  before(:all) { FileUtils.rm_rf(destination) }
+  before(:all) do
```
parkr: Why have 2 before(:all) blocks?

spec/github-pages/integration_spec.rb
benbalter: This defaults the file to the context, meaning if you're in `jekyll-seo-tag` context, it'll look for `jekyll-seo-tag.html` by default.

spec/github-pages/integration_spec.rb
benbalter: No reason. Combined.

Issue #347, opened 2016-11-03
Proposal: add primer/jekyll-octicons


d3v1an7:
### Before submitting an issue, please be sure to

- [x] Read [the contributing instructions](https://github.com/github/pages-gem/blob/master/CONTRIBUTING.md#contributing-to-the-github-pages-gem)
- [x] Update to the latest Gem version (either `gem update github-pages` or `bundle update github-pages`)

### This issue affects

- [x] The site generated by GitHub Pages
- [ ] Building sites locally

### What did you do (e.g., steps to reproduce)
Added https://github.com/primer/jekyll-octicons to site while developing locally (thanks @jonrohan!)

### What did you expect to happen?
Lovely GitHub svg icons, space magic, etc.

### What happened instead?
This error in [repo]/Settings -> GitHub Pages:
> Your site is having problems building: The tag octicon on line 3 in /_layouts/test.html is not a recognized Liquid tag. For more information, see https://help.github.com/articles/page-build-failed-unknown-tag-error.

I've since discovered that only plugins found in this project are supported by GitHub pages (which makes sense!)
I'm about to fix this by manually digging out the assets I need, but I figure `primer/jekyll-octicons` would be a good plugin to support here, given the context :)

citrusui:
I am curious to know why you can't use Jekyll plugins or themes not officially bundled with GitHub Pages. Maybe it would overload page builds?

DirtyF:
@citrusui For plugins, it's mainly for security reasons and avoid execution of arbitrary malicious Ruby code. Themes are rather new, my guess is that GitHub will allow more themes for Jekyll in the foreseeing future.

ndxbn:
In ["Dependency versions" page](https://pages.github.com/versions/), GitHub Pages seems that doesn't include octicons, but since [this commit on Sep 13 2017](https://github.com/github/pages-gem/commit/0c79d7caf66bf209d08d3ee3829c217716a63479), github-pages gem includes "jekyll-octicons" gem.
I published [a page that use octicons in](https://ndxbn.github.io/blog/2018/08/29/testing-octicons), and it seems good.

May I use octicons in GitHub Pages?

PR #483, opened 2017-09-12
Whitelist jekyll-octicons for everyone

parkr:
Whitelists [jekyll-octicons](https://github.com/primer/jekyll-octicons) for everyone.

But do not include as a runtime dependency because not everyone will find it useful.


ndxbn:
ref #347

PR #499, opened 2017-11-27
Make Jekyll Remote Theme enabled by default

benbalter:
A follow up to https://github.com/github/pages-gem/pull/495, this PR versions Jekyll Remote Theme and makes it enabled by default (but will have no effect unless the key `remote_theme` is set in the user's config).

**Edit**: Cleaner diff: https://github.com/github/pages-gem/pull/499/files?w=1

PR #500, opened 2017-11-28
Configure GFM per Pages

kivikakk:
Configure a `markdown: GFM`-configured (!) Jekyll instance like GitHub Pages does. A local clone of a Jekyll site using the `github-pages` gem can just use `markdown: GFM` in their `_config.yml` and it'll work locally exactly like it does in Pages.

A lot of this is duplicated in the inmate image; after merging this we can remove the special-casing there, such that this is what is actually controlling what happens there!

TODO:

- [x] an actual integration test for `GFM` parser

/cc @github/pages

### DIFF
--- a/spec/github-pages/integration_spec.rb
+++ b/spec/github-pages/integration_spec.rb
@@ -165,10 +165,6 @@ RSpec.describe "Pages Gem Integration spec" do
       end
     end
 
-    context "jekyll-commonmark-ghpages" do
-      # noop: this is tested by other things
-    end
-
     context "jekyll-seo-tag" do
       it "outputs the tag" do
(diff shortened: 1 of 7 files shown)


spec/github-pages/integration_spec.rb
```diff
@@ -149,6 +149,10 @@ def env
     end
   end
 
+  context "jekyll-commonmark-ghpages" do
+    # noop: this is tested by other things
```
benbalter: Once https://github.com/github/pages-gem/pull/499 merges, we should be able to add a smoke test for this fairly easily. The main thing here to test is interactions between plugins/depdenencies.

kivikakk: Aha, thank you for that! Super helpful.

PR #501, opened 2017-11-29
script/bump: enforce being on 'master' branch for releases.

parkr:
This partially reverts https://github.com/github/pages-gem/commit/f2b8252c771f12eee4955fdb2b960a69f251f2a0#diff-f9484f4b1a3a5eb061c5fe9c4d5cc7e2L3

Somehow this got deleted (no explanation in commit message), but I think it's important and valid so I propose we keep it.

### DIFF
--- a/script/bump
+++ b/script/bump
@@ -1,5 +1,10 @@
 #!/bin/bash -ex
 
+(git branch | grep -q '* master') || {
+  echo "Only release from the master branch."
+  exit 1
+}
+
 VERSION_FILE="./lib/github-pages/version.rb"
 OLD_VERSION=$(curl "https://rubygems.org/api/v1/versions/github-pages/latest.json" | jq -r .version)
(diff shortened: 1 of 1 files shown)


parkr:
Perhaps it can be tuned to use a plumbing command, or a file in the .git directory.

script/bump
```diff
@@ -1,4 +1,11 @@
-#!/bin/bash -ex
+#!/bin/bash -e
+
+set -e
```
kivikakk: `#!/bin/bash -e` and `set -e`?

parkr: If you run `bash script/bump`, it won't respect that hashbang in my experience. I'll just use one. :)

kivikakk: Ahhhh, yes, that makes perfect sense!

parkr:
CI failed because I merged the PR and the PR ref went away. Rest assured, this script is not under CI.

Issue #509, opened 2017-12-20
Manually requiring jekyll-remote-theme if and only if a remote_theme is requested in the user config


cetinajero:
### Before submitting an issue, please be sure to

- [x] Read [the contributing instructions](https://github.com/github/pages-gem/blob/master/.github/CONTRIBUTING.md#contributing-to-the-github-pages-gem)
- [x] Update to the latest Gem version (run `bundle update github-pages`)

### This issue affects

- [x] The site generated by GitHub Pages
- [x] Building sites locally

### What did you do (e.g., steps to reproduce)

We are having some side effects from https://github.com/github/pages-gem/pull/499 merged PR that was shipped on `pages-gem` [v171](https://github.com/github/pages-gem/releases/tag/v171). We aren't currently using `jekyll-remote-theme` gem neither including the `remote_theme` key on our `_config.yml` files but even with this... we are having side effects due to dependencies from v171 gem related to `libcurl` and we aren't able to compile our sites on Windows machines.

I had tested v171 and v172 on macOS and works just fine because curl comes right-out-the-box but we have some web designers colleagues (that doesn't understand anything of this) working locally on Windows that aren't able to compile anymore their repos due to this dependency problem.

When they exec `bundle exec jekyll serve` on Windows and the Gemfile has `gem 'github-pages', '170', group: :jekyll_plugins` they generate the site successfully.

When they exec `bundle exec jekyll serve` on Windows and the Gemfile has `gem 'github-pages', '171', group: :jekyll_plugins` Jekyll throws an error demanding `libcurl ` due to the inclusion of `jekyll-remote-theme` enabled by default (PR https://github.com/github/pages-gem/pull/499 on pages-gem [v171](https://github.com/github/pages-gem/releases/tag/v171)), it was tested on Win 7, 8 and 10.

### What did you expect to happen?

Everything goes smooth as always on both platforms at the latest `pages-gem` version and the inclusion of a now-hardcode-enabled-but-not-critical-plugin doesn't breaks Jekyll site generation.

We know that the inclusion of the `jekyll-remote-theme` gem will be very helpful for a lot of folks (actually we now plan to create our on Jekyll-theme gem thanks to the inclusion of `jekyll-remote-theme` to GitHub Pages) but it is also causing this side-effects for others who are on windows and suddenly they aren't able to compile their sources by a plugin dependency that aren't even using.

### What happened instead?

A `libcurl` error related to `jekyll-remote-theme` that comes like this:
```
 Dependency Error:
[...]

PR #510, opened 2017-12-20
Only require remote theme if requested

benbalter:
As documented in https://github.com/github/pages-gem/issues/509.

/cc @cetinajero

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -70,6 +70,7 @@ Metrics/ClassLength:
   Exclude:
     - !ruby/regexp /spec\/.*.rb$/
     - !ruby/regexp /features\/.*.rb$/
+    - lib/github-pages/configuration.rb
 
 Style/PercentLiteralDelimiters:
   PreferredDelimiters:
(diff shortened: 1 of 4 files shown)


cetinajero:
Hi @benbalter 

Thanks for your quick response!

I saw that Travis throw this error:
```
lib/github-pages/configuration.rb:7:3: C: Metrics/ClassLength: Class has too many lines. [106/100]
```

You can see on https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/metrics/class_length.rb#L19 that this error puts `[%<length>d/%<max>d]` at the end of the error, from there we can see that the current `configuration.rb` from this PR has 106 LoCs and `Metrics/ClassLength` has been set to *max: 100*

Can I add a commit to this PR where `Metrics/ClassLength` exclude section have this file excluded, it'll be something like this https://github.com/cetinajero/pages-gem/blob/conditionally-require-remote-theme/.rubocop.yml#L73, or what approach do you prefer?

Thanks in advance @benbalter!

cetinajero:
🎉

PR #519, opened 2018-01-23
Optionally add jekyll-remote-theme to the plugins list

benbalter:
A follow up to https://github.com/github/pages-gem/pull/510, this moves the logic to `effective_user_config`, achieving the same result, but earlier, so we don't need custom require statements when we call `set!`.

### DIFF
--- a/lib/github-pages/configuration.rb
+++ b/lib/github-pages/configuration.rb
@@ -108,6 +108,9 @@ module GitHubPages
         # Ensure we have those gems we want.
         config["plugins"] = Array(config["plugins"]) | DEFAULT_PLUGINS
 
+        # To minimize erorrs, lazy-require jekyll-remote-theme if requested by the user
+        config["plugins"].push("jekyll-remote-theme") if config.key? "remote_theme"
+
         if disable_whitelist?
           config["whitelist"] = config["whitelist"] | config["plugins"]
         end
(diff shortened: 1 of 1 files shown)


lib/github-pages/configuration.rb
```diff
@@ -162,15 +164,6 @@ def debug_print_versions
         Jekyll.logger.debug "GitHub Pages:", "github-pages v#{GitHubPages::VERSION}"
         Jekyll.logger.debug "GitHub Pages:", "jekyll v#{Jekyll::VERSION}"
       end
-
-      # To minimize erorrs, lazy-load plugins if they're requested by the user
-      def conditionally_require_plugins(site)
-        plugins = []
-        plugins.push "jekyll-remote-theme" if site.config.key? "remote_theme"
-        Jekyll::External.require_with_graceful_fail(
-          plugins.select { |plugin| site.plugin_manager.plugin_allowed?(plugin) }
```
parkr: Is it ok to remove this? This will change behaviour.

benbalter: > This will change behaviour.

How so? It will be in the config, as if the user had explicitly added it (or it was a default plugin, as it was before), instead of being manually required after the site's config is built.

parkr: Totally right. Don't mind me 🎶

PR #545, opened 2018-03-19
Allow the kramdown math engine to be overridden

stevenkaras:
This will allow the use of the nil math engine in kramdown, which
provides a saner default for templates that don't want mathjax but want
a safe fallback for when javascript is disabled.

From the history of this repo, the math engine used to be changeable up until 2 years ago when kramdown became the primary/only markdown engine. I have not checked if the version of kramdown used back then even had multiple math engines.

### DIFF
--- a/lib/github-pages/configuration.rb
+++ b/lib/github-pages/configuration.rb
@@ -22,6 +22,7 @@ module GitHubPages
         "input"     => "GFM",
         "hard_wrap" => false,
         "gfm_quirks" => "paragraph_end",
+        "math_engine" => "mathjax",
       },
       "exclude" => ["CNAME"],
     }.freeze
@@ -51,7 +52,6 @@ module GitHubPages
       "highlighter" => "rouge",
       "kramdown"    => {
         "template"           => "",
-        "math_engine"        => "mathjax",
         "syntax_highlighter" => "rouge",
       },
       "gist"        => {


aterenin:
+1. Please allow this the math engine to be disabled completely. I use KaTeX, which is lighter and substantially preferable than MathJax for my site.

parkr:
Hey! I'd be happy to accept this PR with the following condition: we override any non-nil math engine to `mathjax`. We cannot permit server-side processing of kramdown's math engines at this time, so setting `math_engine: katex` should be overwritten to `mathjax`. If `math_engine: null`, then we can leave it.

stevenkaras:
The [docs][kramdown-katex] seem to indicate that the KaTeX engine is intended for untrusted user input. Of course, the PR as it stands now would also allow the `sskatex` engine, which is explicitly insecure.

[kramdown-katex]: https://kramdown.gettalong.org/math_engine/katex.html

Perhaps a better way would be to list permitted options, with the first as a default? That would be a more flexible solution moving forward, and keep the current declarative style.

mahrud:
@stevenkaras, any progress on this?

I also prefer allowing `katex` as an option so clients don't need to have javascript enabled, but if it requires a lengthy audit of that gem by GitHub, I'd take the `nil`-or-`mathjax` solution to no solution at all.

parkr:
I don't think we can do any security auditing of new gems, so I'd like to simply allow you to unset the math engine, i.e. allowed values are `nil` and `mathjax`. This is due to amount of time that goes into auditing these new gems for untrusted user input. We have to do our due diligence to keep our users safe, and we can't commit to that at this time.

Allowing the `math_engine` to be `nil` is fine by me.

mahrud:
That sounds reasonable. I don't have access to this PR, so opened a new one off of master with your requested changes.

Issue #482, opened 2017-09-12
Remove auto-generated/injected anchorjs (/assets/javascript/anchor-js)


laukstein:
Issue discussed in https://github.com/jekyll/jekyll/issues/6358

When used [pages-gem recommended](https://github.com/github/pages-gem#usage) **Gemfile**

    gem "github-pages", group: :jekyll_plugins

even when this **Gemfile** and minimal **_config.yml** are the only files in Jekyll dir, on `jekyll serve` it will generate **_site/assets** with [anchorjs](https://github.com/bryanbraun/anchorjs). Why?

![grabilla g13256](https://user-images.githubusercontent.com/183966/30325522-bfeeb4a8-97cd-11e7-8146-ef7c24095a57.png)

The only way I found to remove **_site/assets** was to remove **Gemfile** or have it `gem "github-pages"` without ~`group: :jekyll_plugins`~.

benbalter:
@laukstein I believe if you set `theme: none` or `theme: ""` it should not use the Primer theme (which is the source of the asset).

laukstein:
@benbalter, Jekyll fails when used `theme: none` or `theme: ""` in **Gemfile**, see

with `theme: none`

> Configuration file: .../_config.yml
> jekyll 3.5.2 | Error:  The none theme could not be found.

with `theme: ""`

> Configuration file: ../_config.yml
> jekyll 3.5.2 | Error:  The  theme could not be found.

laukstein:
@mattr, it gets me generated without `theme` key. See the above comment, the issue might be related to `group: :jekyll_plugins`.

mattr-:
I don't put the `github-pages` gem in the `jekyll-plugins` group either, so maybe that's the ultimate fix? No idea.

laukstein:
@mattr-, [github/pages-gem](https://github.com/github/pages-gem) recommendation is to use `group: :jekyll_plugins`, so #482 is a issue.

benbalter:
@mattr- https://github.com/pages-themes/primer is the default theme, set by the GitHub Pages Gem when no other theme is specified, so that workflows can be as simple as committing markdown files and having a site. It is largely independent, unless of course, you care about extraneous assets i the built site.

@laukstein What if you add the following to your config (and remove the theme suggestion above):

```yml
exclude:
  - assets/javascript/anchor-js
```

I believe that should exclude the extraneous assets from appearing in your built site.

laukstein:
@benbalter,

>     exclude:
[...]

PR #553, opened 2018-04-15
Allow theme to be explicitly disabled

ntkme:
This PR allows theme to be explicitly disabled via `theme: null`.

When a source site contains its theme, files from the default theme `jekyll-theme-primer` will still ends up in the dest site. The most annoying part of this issue (pages-themes/primer#15) is that `assets/css/style.css` from `jekyll-theme-primer` would show up in `site.pages` collection (e.g. `sitemap.xml`). The upstream seems to have no interest in fixing that issue, and I don't really care either as long as I can avoid that buggy theme.

According to [Jekyll::Site.configure_theme](https://github.com/jekyll/jekyll/blob/29dc190fcb41f13dfe9764d1d539a34657d44788/lib/jekyll/site.rb#L435), supposedly a user should be able to disable any theme by having `theme: null` in `_config.yml`. However, due to the way configurations are merged in [GitHubPages::Configuration.effective_config](https://github.com/github/pages-gem/blob/938b95356d0dda654d2323ff7075b38b20ce597b/lib/github-pages/configuration.rb#L100) (eventually the logic in [Jekyll::Utils.deep_merge_hashes](https://github.com/jekyll/jekyll/blob/29dc190fcb41f13dfe9764d1d539a34657d44788/lib/jekyll/utils.rb#L319-L320)), an explicit `theme: null` will still be overrode by `theme: "jekyll-theme-primer"`.

Changing the behavior of `Jekyll::Utils.deep_merge_hashes` on explicit `null` in yaml (or `nil` in ruby) is unrealistic, as it is so fundamental that tons of existing code depend on its current behavior. Thus, the best solution is fixing it the after configuration merging in this repository.

#482 mentioned using a hack of `theme: false`, it works because [Jekyll::Site.configure_theme](https://github.com/jekyll/jekyll/blob/29dc190fcb41f13dfe9764d1d539a34657d44788/lib/jekyll/site.rb#L441-L443) treats a boolean `false` as an invalid value and then fallback to `nil`. This PR is intended to avoid the such hack and properly allow theme to be explicitly disabled.

### DIFF
--- a/lib/github-pages/configuration.rb
+++ b/lib/github-pages/configuration.rb
@@ -101,6 +101,9 @@ module GitHubPages
           .fix_common_issues
           .add_default_collections
 
+        # Allow theme to be explicitly disabled via "theme: null"
+        config["theme"] = user_config["theme"] if user_config.key?("theme")
+
         exclude_cname(config)
 
         # Merge overwrites into user config
--- a/spec/github-pages/configuration_spec.rb
+++ b/spec/github-pages/configuration_spec.rb
@@ -118,6 +118,17 @@ describe(GitHubPages::Configuration) do
         end
       end
 
+      context "with user-specified theme to be null" do
+        let(:site) do
+          config = configuration.merge("theme" => nil)
+          Jekyll::Site.new(config)
+        end
+
+        it "respects null" do
+          expect(site.theme).to be_nil
+        end
+      end
+
       it "plugins don't include jekyll remote theme" do
         expect(effective_config["plugins"]).to_not include("jekyll-remote-theme")
       end


ntkme:
Test is failing on ruby 2.5 due to travis-ci/travis-ci#8969.

lib/github-pages/configuration.rb
```diff
@@ -101,6 +101,11 @@ def effective_config(user_config)
           .fix_common_issues
           .add_default_collections
 
+        # Allow theme to be explicitly disabled via "theme: null"
+        if user_config.key?("theme") && user_config["theme"].nil?
+          config["theme"] = nil
```
parkr: What do you think about something like:

```text
config["theme"] = user_config["theme"] if user_config.key?("theme")
```

ntkme:
@parkr Your version is more readable and less confusing, so I've happily followed your advise.

ntkme:
@parkr @benbalter Any update?

MikeMcQuaid:
Thanks @ntkme!

PR #563, opened 2018-05-14
Require nokogiri 1.8.2 or greater

parkr:
Addresses CVE-2017-18258
https://nvd.nist.gov/vuln/detail/CVE-2017-18258

This prevents folks from using a nokogiri version (< 1.8.2) affected
by the above vulnerability.


.ruby-version
```diff
@@ -1 +1 @@
-2.4.0
+2.2.10
```
benbalter: ? Assume this is unrelated? Should we remove the constraint entirely or is this what's used in Production?

PR #569, opened 2018-06-19
fixes misspelled word

kenman345:
Always fun when you get to correct an error in spelling error.... :-D


benbalter:
Thanks @kenman345!

kenman345:
Always happy to help. I read through documentation way too often and find errors not to try and fix it when I see it.

Issue #589, opened 2018-09-06
Gemfile is hosted in GitHub pages


ahuglajbclajep:
### Before submitting an issue, please be sure to

- [x] Read [the contributing instructions](https://github.com/github/pages-gem/blob/master/docs/CONTRIBUTING.md#contributing-to-the-github-pages-gem)
- [x] Update to the latest Gem version (run `bundle update github-pages`)

### This issue affects

- [x] The site generated by GitHub Pages
- [ ] Building sites locally

*The GitHub Pages Gem is intended to help users replicate the GitHub Pages build environment locally. If your issue affects both the hosted version and the version previewed locally, you may be better suited reporting [seeking support in other forums](https://github.com/github/pages-gem/blob/master/docs/SUPPORT.md).*

### What did you do (e.g., steps to reproduce)
I pushed some markdown files and Gemfile to my repository in GitHub.

### What did you expect to happen?
According to the [default configuration](https://jekyllrb.com/docs/configuration/default/), Gemfile will not be hosted.

### What happened instead?
Gemfile is hosted.

### Additional information
In local build(`bundle exec jekyll b`), Gemfile is not include in *_site* directory.
I think that Gemfile was not hosted before.
Is this a specification?

* Link to the live site (if applicable): https://prog-g.github.io/Gemfile
* Link to the source repo (if applicable): https://github.com/prog-g/prog-g.github.io/tree/cbed726523e2ff74b06afa8d5e6a1abe59b9c6e3

ahuglajbclajep:
> I think that Gemfile was not hosted before.

I'm sorry, I misunderstood.
Gemfile was hosted since before.

ahuglajbclajep:
If you don't want to host Gemfile, you should write the following in `_config.yml`.
```yml
exclude:
  - Gemfile
```
This override the default config, so please exclude `node_modules` or `vendor` if necessary.

PR #602, opened 2018-11-22
Add test to indicate Jekyll default excludes are retained

mrmanc:
This change adds an assertion to indicate that the default Jekyll
exclude configuration is merged into the configuration set by the
pages-gem configuration. The defaults can be seen on the [Jekyll Site]
(https://jekyllrb.com/docs/configuration/default/).

After encountering the same issue described in #589 I looked at whether
the default excludes were preserved by the pages-gem, and was confused
by the difference in the behaviour I observed vs what the code said. So
I wrote this test to indicate that the code really does work as you’d
hope it did.

### DIFF
--- a/spec/github-pages/configuration_spec.rb
+++ b/spec/github-pages/configuration_spec.rb
@@ -62,6 +62,10 @@ describe(GitHubPages::Configuration) do
       expect(effective_config["exclude"]).to include("CNAME")
     end
 
+    it "retains Jekyll default excludes" do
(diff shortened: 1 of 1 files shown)


parkr:
@mrmanc Hey! I looked into this yesterday in production and a corresponding test fails. The reason is that we're not using `Jekyll.configuration` in production – we're reading and merging somewhat manually. I wrote #607 to match the two.

mrmanc:
Awesome, thanks for merging and continuing to address the issue 👍

parkr:
@mrmanc I went ahead and shipped a fix internally – please let me know if it's not working for you!

PR #607, opened 2018-12-13
Configuration: read the config files manually to better mimic the production environment

parkr:
In GitHub Pages, we manually read the configuration files instead of relying on `Jekyll.configuration` like Jekyll does. This leads to some confusing behavior between local Jekyll runs and production GitHub Pages runs.

This PR moves the reading logic into this gem which the production system can then rely on.

The key parts:

1. Added a class called `GitHubPages::ConfigurationReader` whose sole purpose is to read configuration files and merge them together (no defaults or anything applied!)
2. Use the ConfigurationReader in the tests so we can match production (and use it in production too!)
3. Ensure things like `DEFAULTS["exclude"]` are a concatenation of our settings instead of overwriting
4. Update tests to try to allow for easier config overrides.

### DIFF
--- a/lib/github-pages.rb
+++ b/lib/github-pages.rb
@@ -5,10 +5,11 @@ require "jekyll"
 
 # Top-level namespace for all GitHub Pages-related concerns.
 module GitHubPages
-  autoload :Plugins,       "github-pages/plugins"
-  autoload :Configuration, "github-pages/configuration"
-  autoload :Dependencies,  "github-pages/dependencies"
-  autoload :VERSION,       "github-pages/version"
+  autoload :Plugins,             "github-pages/plugins"
+  autoload :Configuration,       "github-pages/configuration"
+  autoload :ConfigurationReader, "github-pages/configuration_reader"
+  autoload :Dependencies,        "github-pages/dependencies"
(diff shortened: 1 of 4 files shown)


parkr:
@benbalter Apologies for the ping – I'm requesting your review here since you have the most experience with this section of both codebases besides myself ❤️ ❤️ ❤️

lib/github-pages.rb
```diff
@@ -5,10 +5,11 @@
 
 # Top-level namespace for all GitHub Pages-related concerns.
 module GitHubPages
-  autoload :Plugins,       "github-pages/plugins"
-  autoload :Configuration, "github-pages/configuration"
-  autoload :Dependencies,  "github-pages/dependencies"
-  autoload :VERSION,       "github-pages/version"
+  autoload :Plugins,             "github-pages/plugins"
```
AniDubEA, line 8:
```suggestion
  autoload :Plugins,             "github-pages/plugins"
```

AniDubEA (commented):
#

> 7

> 8

> 9

> 10

> 11

> 12

AniDubEA (approved):

PR #623, opened 2019-02-15
Bump jemoji to 0.10.2

janko:
That version resolves broken emoji links after GitHub updating its assert host. See https://github.com/jekyll/jemoji/issues/89.


chrisarusso:
We'd love to have our emoji back and working on our production site ASAP. Thanks.

PR #643, opened 2019-05-31
Allow jekyll-archives plugin

RDIL:
Its an official plugin and I want it on my site


RDIL:
@parkr even though you guys need to audit gems for security, I feel like it is trustworthy as it is officially made by Jekyll.

PR #655, opened 2019-10-10
Un-pin activesupport

### LINKED ISSUE #654: Relax Active Support dependency
This line locks ActiveSupport to 4.2.1.1:

https://github.com/github/pages-gem/blob/2747fa68f5248cb40be9fd272a039c2fad0e2ac7/lib/github-pages/dependencies.rb#L48

This 4.2.x branch of Rails started in 2014, and so this dependency is locked to a 5 year old branch.

Could we please have this dependency relaxed so that we can use any modern-ish version of ActiveSupport?

parkr:
This allows users to specify their preferred activesupport version within the confines of the other dependencies.

Fixes #654.


parkr:
Merging this with https://github.com/github/pages-gem/pull/664.

PR #660, opened 2019-10-30
Upgrade rubocop to v0.76.0

parkr:
The Rubocop version here was very old.


benbalter:
@parkr thoughts on moving things over to https://github.com/github/rubocop-github now that it's a thing? Glad to push up a commit if you'd like?

parkr:
@benbalter Bummer. If you'd like to start a revert PR for this PR and give rubocop-github a try, I think using a centralized rubocop gem for the organization makes more sense than trying to be bespoke.

PR #691, opened 2020-06-05
Allow jekyll-loading-lazy plugin

gildesmarais:
Making the generated sites load faster by utilizing native lazy loading
with [jekyll-loading-lazy](https://github.com/gildesmarais/jekyll-loading-lazy).

The change regarding the expected image url in jekyll-avatar spec
stems from  Nokogiri's handling of urls (which is basically done on
libxml2's end). See
https://github.com/sparklemotion/nokogiri/issues/1127#issuecomment-453737666
for more details on that topic. The encoding of the ampersand in the
search params of the image url does not cause the url to break or point
to an invalid location.

I'd love to see my Jekyll plugin become allowed and ready to
use on Github Pages. Thank you!


solrevdev:
+1

gildesmarais:
After 1.5 years without receiving any attention, I feel this addition isn't placed high on the priority list.

PR #704, opened 2020-08-05
Bump jekyll to v3.9.0, kramdown to v2.3.0

hubot:
**Release**: https://github.com/jekyll/jekyll/releases/tag/v3.9.0
**Diff**: https://github.com/jekyll/jekyll/compare/v3.8.7...v3.9.0


ashmaroli:
Shouldn't GHP whitelist the remaining kramdown extensions so that users can opt-in to use them if needed?
In the current state, a user using, say the Coderay highlighter from kramdown v1 will suddenly have lost their code highlighting and will further not be able to load the `kramdown-syntax-coderay` gem to resolve the issue either.

parkr:
> Shouldn't GHP whitelist the remaining kramdown extensions so that users can opt-in to use them if needed?
> In the current state, a user using, say the Coderay highlighter from kramdown v1 will suddenly have lost their code highlighting and will further not be able to load the `kramdown-syntax-coderay` gem to resolve the issue either.

I'm not sure that's true. The PLUGIN_WHITELIST variable is passed into Jekyll as the `whitelist` config entry. This is only used in the PluginManager. Since the KramdownParser code does the `require`-ing itself, the gems do not need to be part of this `whitelist` variable.

ashmaroli:
I see. Thanks for clearing it for me. :+1:

ashmaroli:
One more question: What if a user would like to continue using kramdown v1 on GitHub Pages?

parkr:
This PR updates Math output from `<script type='math/tex; mode=display'>` to `\[ ... \]`. Please make sure your JavaScript library is updated to support `\[ ... \]`.

For context: https://github.com/gettalong/kramdown/issues/626 https://github.com/gettalong/kramdown/issues/651

PR #772, opened 2021-04-26
Remove gem-based themes, replace with remote themes

parkr:
This updates the github-pages gem to remove the gem-based themes. These are quite a burden to keep up-to-date. It's much better to use jekyll-remote-theme which pulls down the theme on-demand. Folks can specify a new version if they like with `remote_theme: user/repo@revision`. I think this will be more maintainable and it will reduce the number of dependencies for `github-pages`.


YiMysty (approved):

0xf10yd (approved):

74971030177:
I just won't be needing these

PR #798, opened 2021-11-15
Add build functionality to `github-pages` executable

jriggins:
Adds the `build` option to the `github-pages` executable and add CI to build and publish Docker images (publish only on `main` and `v*` tagged branches)

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -1,0 +1,35 @@
+name: Publish to Docker
+on:
+  push:
+    branches: [ master ]
+    tags: [ 'v*' ]
+  pull_request:
+    branches: [ master ]
+ 
+env:
+  DOCKER_TAG_IMAGE: "ghcr.io/github/pages-gem"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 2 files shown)


felipesu19 (approved):

.github/workflows/publish-docker.yml
```diff
@@ -0,0 +1,35 @@
...
+        uses: docker/metadata-action@v3
+        with:
+          images: ${{ env.DOCKER_TAG_IMAGE }}
+
+      - name: Build Docker Image
+        run: |
+          docker build -t ${{ steps.meta.outputs.tags }} -f Dockerfile .
+
+      - name: Push to Container Registry
+        if: github.event_name != 'pull_request'
```
jcudit, line 31: Should we only run this job on `push` if we discard the build for the `pull_request` event?

jcudit (approved):

.github/workflows/publish-docker.yml
jriggins: Discussed out of band, but I think we settled on continuing to allow this to run on PRs but only publishing to the package registry on pushes to main and tagged branches.

Issue #810, opened 2022-01-02
Versions out of sync


TWiStErRob:
### Before submitting an issue, please be sure to

- [x] Read [the contributing instructions](https://github.com/github/pages-gem/blob/master/docs/CONTRIBUTING.md#contributing-to-the-github-pages-gem)
- [x] Update to the latest Gem version (run `bundle update github-pages`)

### This issue affects

- [x] The site generated by GitHub Pages
- [x] Building sites locally

> The GitHub Pages Gem is intended to help users replicate the GitHub Pages build environment locally.

### What did you do (e.g., steps to reproduce)
Want to update my `Gemfile` to match GitHub's version.

### What did you expect to happen?
Find the specific version GitHub uses at the moment.

### What happened instead?
Confusion:
 * https://rubygems.org/gems/github-pages lists 219..223 (and older behind link)
 * https://github.com/github/pages-gem#github-pages-ruby-gem has a badge for 223
 * https://github.com/github/pages-gem/releases shows 221
 * https://pages.github.com/versions/ shows 219

At this point I can't tell if the `/versions/` page is outdated (August?!) or GitHub is not using the latest from its own project?!

Please clarify and update scripts/documentation to keep these 4 numbers in sync.

TWiStErRob:
Note: not having 222 and 223 on GitHub releases also means it's hard to see what's changed.

parkr:
@yoannchaudet Hey! Is there a way to run a Page Build job on the pages.github.com site periodically? Maybe a GitHub Action could fire off the build? Many folks use https://pages.github.com/versions/ (and `/versions.json`) to understand what gems they should use to match GitHub Pages, so ideally the site is rebuilt whenever you launch a new version to Pages. Thank you!

yoannchaudet:
Hello here!

Thank you for pointing these inconsistencies, this was not intended.

Versions 222 and 223 did not include any changes relevant to the build process itself and just helped support https://github.blog/changelog/2021-12-16-github-pages-using-github-actions-for-builds-and-deployments-for-public-repositories/.

We should not be out of sync in so many locations! I'll make sure we update our procedure to release `pages-gem` so it does not happen again and so we clean stuff up in the next few days.

@parkr (hi!) yes! We can definitely automate the rebuilds for pages.github.com ❤️

parkr:
@yoannchaudet Hi! Thank you so much! ❤️

TWiStErRob:
Thanks @parkr for pointing to the right person!

[...]

PR #812, opened 2022-01-12
Update automation

felipesu19:
This change seeks to ensure we mantain gem version parity as per [this issue ](https://github.com/github/pages-gem/issues/810) and the [subsequent card ](https://github.com/github/pages-engineering/issues/988), we really need to ensure we update the gem version in at least three places. 

This is the first pass at automating that change, via adding an automated tag release via the gh cli and creating new prs in the relevant repos so we know what work needs to be done, which will fire whenever anyone runs the release script

As part of the cleanup task I'd like to 

1. Add automation to the relevant repos that detects the new pr after the script creates them, runs an action and makes the changes based on that 
2. Automatically commits that change to the pr and sends a message to the pr creator. This should reduce the burden to just having to run three deploys (two actually, based on point 3)
3. I'd also like to get pages.github.com to somehow listen to changes via a github action, rather than rebuilding itself every 6 hours as originally suggested. However if that's too finicky (and based on preliminary investigation it may well be), we should just get it auto-rebuilding. If we do that we can cut the part of the automation that concerns itself with that repo. 
4. Finally, the automation could use some hardening, the script as it exists has no retry logic and no real failure handling. 

This is ready for a first pass. I've manually tested this by running through the command flow starting with line 46 in the script/release file (with some manual interpolations), as part of the work I'm going to create a new release which will test things end to end and let me clean up the dependent repos, but i'm going to tackle #3 first. 

Feedback on the idea, the implementation, and the general direction of this work. Its possible that point #1 is too much work for how often we release new gems, and we should focus on 2 and 3.

### DIFF
--- a/script/release
+++ b/script/release
@@ -82,9 +82,9 @@ gh api repos/actions/jekyll-build-pages/git/refs/heads/pages-gem-release-$tag -F
 gh api repos/github/pages.github.com/git/refs/heads/pages-gem-release-$tag -F "sha"="$new_gh_pages_site_sha" -F "ref"="refs/heads/pages-gem-release-$tag" -F "force"="true"
 
 # create pull requests in downstream repos
-gh pr create --base "master" --reviewer "@me" --head "pages-gem-release-$tag" --title "Bump pages-gem version to $tag" --assignee "@me" --body "Bump pages-gem version to $tag. You must manually make those changes and commit to this branch" --repo "github/pages" 
-gh pr create --base "main" --reviewer "@me" --head "pages-gem-release-$tag" --title "Bump pages-gem version to $tag" --assignee "@me" --body "Bump pages-gem version to $tag. You must manually make those changes and commit to this branch" --repo "actions/jekyll-build-pages"
-gh pr create --base "master" --reviewer "@me"  --head "pages-gem-release-$tag" --title "Bump pages-gem version to $tag" --assignee "@me" --body "Bump pages-gem version to $tag. You must manually make those changes and commit to this branch" --repo "github/pages.github.com"
+gh pr create --base "master" --head "pages-gem-release-$tag" --title "Bump pages-gem version to $tag" --assignee "@me" --body "Bump pages-gem version to $tag. You must manually make those changes and commit to this branch" --repo "github/pages" 
+gh pr create --base "main"  --head "pages-gem-release-$tag" --title "Bump pages-gem version to $tag" --assignee "@me" --body "Bump pages-gem version to $tag. You must manually make those changes and commit to this branch" --repo "actions/jekyll-build-pages"
+gh pr create --base "master" --head "pages-gem-release-$tag" --title "Bump pages-gem version to $tag" --assignee "@me" --body "Bump pages-gem version to $tag. You must manually make those changes and commit to this branch" --repo "github/pages.github.com"
 
 # clean up the input file for the commit
 
--- a/script/release
+++ b/script/release
@@ -1,4 +1,9 @@
 #!/bin/sh
+
+# check that gh is installed and logged in 
+
+gh auth status -h github.com
+
 # Tag and push a release.
 
 set -e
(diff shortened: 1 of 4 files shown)


YiMysty (approved):

script/release
```diff
@@ -39,4 +39,53 @@ git fetch -t origin
 # Tag it and bag it.
 
 gem push github-pages-*.gem && git tag "$tag" &&
-  git push origin master && git push origin "$tag"
+git push origin master && git push origin "$tag"
+
+# create a new release on github
+gh release create "$tag" --notes "Automated release for $tag"
```
tcbyrd, line 45: Is this assuming `gh` is authenticated locally? I think this is a new dependency in the scripts, so we might want to ensure it's documented and verify that `gh` has the appropriate permissions in the script. We can do something like [`gh auth status -h github.com`](https://cli.github.com/manual/gh_auth_status) to verify it's logged in and credentialed for `github.com` before continuing.

script/release
```diff
@@ -39,4 +39,53 @@ git fetch -t origin
...
+
+gh api repos/github/pages/git/refs -F "sha"="$gh_pages_sha" -F "ref"="refs/heads/pages-gem-release-$tag"
+gh api repos/github/pages/git/refs -F "sha"="$jekyll_sha" -F "ref"="refs/heads/pages-gem-release-$tag"
+gh api repos/github/pages/git/refs -F "sha"="$gh_pages_site_sha" -F "ref"="refs/heads/pages-gem-release-$tag"
+
+#then we need the tree object so we can create a commit using the api
+
+gh_pages_tree = gh api repos/github/pages/git/commits/$gh_pages_sha | jq '. | .tree.sha'
+jekyll_tree = gh api repos/github/actions/jekyll-build-pages/commits/$jekyll_sha | jq '. | .tree.sha'
+gh_pages_site_tree = gh api repos/github/papages.github.comges/git/commits/$gh_pages_site_sha | jq '. | .tree.sha'
```
tcbyrd, line 64:
```suggestion
gh_pages_site_tree = gh api repos/github/pages.github.com/git/commits/$gh_pages_site_sha | jq '. | .tree.sha'
```

script/release
```diff
@@ -39,4 +39,53 @@ git fetch -t origin
...
+
+# create the input file for the commit. We have to do this because the F flag doesn't support arrays
+
+echo "{\"tree\": \"$gh_pages_tree\", \"parents\": [\"$gh_pages_sha\"], \"message\": \"automated commit for pr creation\"}" > gh_pages_commit_input.json
```
tcbyrd, line 68: Wonder if we should put these into `/tmp` folder that gets gitignored? I know all this happens after the commit flow, but if it errors out it could leave responses as untracked files in the repo.

felipesu19: I like it, will update the flow to do that

felipesu19:
Gonna merge this in, will open a new pr with the tagging bit of the work

script/release
```diff
@@ -1,4 +1,9 @@
 #!/bin/sh
+
+# check that gh is installed and logged in 
+
+gh auth status -h github.com
```
parkr, line 5: Since this is before the `set -e`, if it fails it won't stop the script from running.

script/release
```diff
@@ -39,4 +44,50 @@ git fetch -t origin
...
+
+
+# create a tmp folder
+mkdir -p tmp
+
+#for each dependent repo, create a new branch an a blank pr. 
+# first get the sha of main for each repo 
+
+gh_pages_sha = gh api repos/github/pages/git/refs/heads/master | jq '. | .object.sha'
+jekyll_sha = gh api repos/actions/jekyll-build-pages/git/refs/heads/main | jq '. | .object.sha'
```
parkr, line 60: What shell are you using? I thought you'd need backticks to execute commands and return their output.

parkr (commented):
Some quick comments for your next PR!

PR #829, opened 2022-04-08
build: don't ignore Gemfile.lock for deterministic builds

dijonkitchen:
Reference:
https://bundler.io/rationale.html#checking-your-code-into-version-control

Seems to match what is published: https://pages.github.com/versions/


fulldecent:
This PR seems to make a failed assertion.

When GitHub Pages builds a website (not talking about GitHub Actions) it does NOT use the Gemfile.lock to find what versions to build with.

It means that GitHub Pages' infrastructure decides which versions of Jekyll/Ruby/... it is running, not you.

This PR does not change that behavior of how GitHub Pages infrastructure works.

So if this is PR is accepted, then the test suite will run against new versions of Jekyll/Ruby/... rather than the version that production uses.

If my understand is correct here, we should NOT merge this PR.

parkr:
RubyGems do not ship with their Gemfile.lock files. The guidance you linked to is for end users. There is no way for an end user to use the Gemfile.lock from a RubyGem.

Issue #610, opened 2018-12-21
Whitelist jekyll-data plugin


yatil:
In lieu of a proper translation solution (#401), can we at least make data in plugins portable by whitelisting https://github.com/ashmaroli/jekyll-data. Keeping data that belongs to the theme (like common navigation or translation phrases) in sync is really complicated at the moment. This would be a relatively simple solution.

big-andy-coates:
Yep, this would be really useful, e.g. to maintain a consistent navigation menu when a site is built by composing the gh pages of multiple gh repos.

Currently, I have to keep data files synchronised across 20 odd repos.  A right PITA! ;)

big-andy-coates:
Related to https://github.com/benbalter/jekyll-remote-theme/issues/68

parkr:
We upstreamed theme data folder propagation to Jekyll: https://github.com/jekyll/jekyll/pull/8815

I believe this repo is in maintenance mode. GitHub Actions now has first-class support for Pages, so I'd recommend setting that up for affected repos if you can.

parkr:
Use Jekyll v4.3.0 or later. Full release notes (including the above linked PR): https://github.com/jekyll/jekyll/releases/tag/v4.3.0

big-andy-coates:
Thanks!

big-andy-coates:
A bit of Googling on this seems to suggest that this involves switching to a workflow to build the docs, and it may well be the case that the thème I'm using isn't compatible with Jekyll 4.x.

I'm therefore wondering if this whitelisting would enable the required functionally without the need to switch to workflow or finding a Jekyll 4.x theme.

Thoughts @parkr ?

parkr:
@big-andy-coates The Jekyll v4 upgrade isn't that big actually – the reason I never got around to it while I worked at GitHub was because there was a breaking change in `post_url` which would have silently broken millions of sites. But if you're doing the upgrade yourself, you can check on links throughout the site to ensure they're correct. Jekyll even has upgrading docs: https://jekyllrb.com/docs/upgrading/3-to-4/

I don't work for GH anymore but from what I gather on their blog, Actions is The Future ™️  so they're likely not accommodating the security review and maintenance of additional plugins at this time. I can't say for sure though, but if I were you, I'd migrate to an (Actions) workflow and have full customizability. :)

PR #855, opened 2022-11-12
Whitelist the `jekyll-data` plugin

big-andy-coates:
fixes: https://github.com/github/pages-gem/issues/610

Whitelist [jekyll-data](https://github.com/ashmaroli/jekyll-data).

This allows people using GitHub pages to use themes that provide defaults in their `_config.yml` and have data under `_data/`.

For example, this would allow organisations that compose multiple GitHub repositories into a single site to use a common theme, that sets defaults in the theme's `_config.yml` and has shared `_data` structures, like navigation menus,

### DIFF
--- a/lib/github-pages/plugins.rb
+++ b/lib/github-pages/plugins.rb
@@ -39,6 +39,7 @@ module GitHubPages
       jekyll-include-cache
       jekyll-octicons
       jekyll-remote-theme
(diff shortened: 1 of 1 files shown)


big-andy-coates:
Release to: https://github.com/benbalter/jekyll-remote-theme/issues/68

PR #865, opened 2023-01-20
Fix compatibility with new ruby versions

domoritz:
Update liquid because of https://github.com/jekyll/jekyll/issues/9233.


yoannchaudet:
This landed earlier!

PR #871, opened 2023-02-07
Bump rouge to v3.30.0

fabiomadge:
Last release of that major version.


yoannchaudet:
This made it in https://github.com/github/pages-gem/pull/902 and was released under v229!

Issue #752, opened 2021-01-11
Jekyll serve fails on Ruby 3.0 (webrick missing)


Krinkle:
### This issue affects

- [ ] The site generated by GitHub Pages
- [x] Building sites locally

### What did you do (e.g., steps to reproduce)

```
bundle exec jekyll serve
```

On macOS 10.15 with Ruby 3.0.0 from Homebrew, using the latest github-pages in a Gemfile.

```
# Gemfile
gem "github-pages"

# Gemfile.lock:
    github-pages (209)
      jekyll (= 3.9.0)
…
```

### What happened instead?

```
bundler: failed to load command: jekyll (/usr/local/lib/ruby/gems/3.0.0/bin/jekyll)
/usr/local/lib/ruby/gems/3.0.0/gems/jekyll-3.9.0/lib/jekyll/commands/serve/servlet.rb:3:in `require': cannot load such file -- webrick (LoadError)
	from /usr/local/lib/ruby/gems/3.0.0/gems/jekyll-3.9.0/lib/jekyll/commands/serve/servlet.rb:3:in `<top (required)>'
	from /usr/local/lib/ruby/gems/3.0.0/gems/jekyll-3.9.0/lib/jekyll/commands/serve.rb:184:in `require_relative'
	from /usr/local/lib/ruby/gems/3.0.0/gems/jekyll-3.9.0/lib/jekyll/commands/serve.rb:184:in `setup'
	from /usr/local/lib/ruby/gems/3.0.0/gems/jekyll-3.9.0/lib/jekyll/commands/serve.rb:102:in `process'
	from /usr/local/lib/ruby/gems/3.0.0/gems/jekyll-3.9.0/lib/jekyll/commands/serve.rb:93:in `block in start'
	from /usr/local/lib/ruby/gems/3.0.0/gems/jekyll-3.9.0/lib/jekyll/commands/serve.rb:93:in `each'
	from /usr/local/lib/ruby/gems/3.0.0/gems/jekyll-3.9.0/lib/jekyll/commands/serve.rb:93:in `start'
	from /usr/local/lib/ruby/gems/3.0.0/gems/jekyll-3.9.0/lib/jekyll/commands/serve.rb:75:in `block (2 levels) in init_with_program'
	from /usr/local/lib/ruby/gems/3.0.0/gems/mercenary-0.3.6/lib/mercenary/command.rb:220:in `block in execute'
[...]

PR #881, opened 2023-04-23
Mention Ruby version support and WEBrick dependency in README.md

solvaholic:
per https://github.com/github/pages-gem/issues/752#issuecomment-764758292:

> Hi! In general, we recommend (and support) running the github-pages gem with the Ruby version we're running in production (2.7.x). We don't yet support Ruby 3.0.0. Ruby 2.7.x still has webrick so I'd recommend using that until we've upgraded for your GitHub Pages needs!


parkr (commented):
I'd vote to add the webrick gem by default to github-pages instead.

janbrasna:
This is resolved as of [v232](https://github.com/github/pages-gem/releases/tag/v232).

PR #882, opened 2023-05-10
Fix link to pages.github.com in README

merwan:
The original link is a 404, fix the destination to https://pages.github.com


YiMysty:
👋 @merwan thanks for making the pr. However the URL is a correct link to a existed private repo, we will not merge this change.

PR #894, opened 2023-12-15
Bump terminal-table to 3.0.2

anna328p:
Prevents a dependency conflict with `steep`.

### DIFF
--- a/github-pages.gemspec
+++ b/github-pages.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency("mercenary", "~> 0.3")
   s.add_dependency("nokogiri", ">= 1.13.6", "< 2.0")
-  s.add_dependency("terminal-table", "~> 1.4")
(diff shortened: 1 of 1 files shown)


jabenninghoff:
+1 on this PR, this is also causing dependency conflicts with the version of `rubocop` needed for the currently used version of ruby (3.3.4). Can this be merged?

github-pages.gemspec
```diff
@@ -26,7 +26,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency("mercenary", "~> 0.3")
   s.add_dependency("nokogiri", ">= 1.13.6", "< 2.0")
-  s.add_dependency("terminal-table", "~> 1.4")
+  s.add_dependency("terminal-table", "~> 3.0")
```
parkr, line 29: We might prefer something like >= 1.4, <4 for now unless our use is broken between these versions?

anna328p: Will update.

PR #897, opened 2023-12-28
Bump jekyll to v3.9.4

parkr:
Adds Ruby 3.3 support. (https://github.com/jekyll/jekyll/releases/tag/v3.9.4)

@yoannchaudet, if you'd like to bring support up to Ruby 3.3, this tiny v3.9.4 release has been made available for you!

PR #898, opened 2024-01-02
Create publish-gem.yml

tsusdere:
Adds workflow to automate publishing the gem to RubyGems


.github/workflows/publish-gem.yml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '3.2'
```
YiMysty, line 17:
```suggestion
          ruby-version: '2.7.3'
```

.github/workflows/publish-gem.yml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '2.7.3'
```
parkr, line 17: Is there any reason not to release on a later version? At least worthwhile to use 2.7.8 to get the bug and security fixes.

YiMysty: 👋 @parkr do you think if we can move to ruby 3.0 for this project? yeh we can move to 2.7.8

.github/workflows/publish-gem.yml
```diff
@@ -0,0 +1,23 @@
...
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '2.7.8'
+      - name: Build gem
+        run: |
+          gem build github-pages.gemspec
+      - name: Publish
```
YiMysty, line 21: this should only run when merge into master branch I think?

tsusdere: Do we want to instead run the job when we merge into `master` branch?
```yaml
on:
  push:
    branches:
      - main
```

YiMysty:
```
on:
  release:
    types: [released]
```

.github/workflows/publish-gem.yml
tsusdere: To add seems the same problem with CI can be seen with other repos (regarding Ruby version):
https://github.com/pages-themes/midnight/actions/runs/7391072383/job/20107215490?pr=50

parkr: I think we should go to 3.2 at least, but there might be bugs to fix along the way.

Dockerfile
```diff
@@ -8,7 +8,8 @@ RUN apt-get update \
     make \
     nodejs
 
-RUN gem update --system
+# final version for rubygems-update to supoprt ruby 2.7, consider removing this after upgraded to ruby 3.0
+RUN gem update --system 3.4.22
```
parkr, line 12: Since you're upgrading in this PR, you can probably remove this.

.github/workflows/push.yml
```diff
@@ -7,8 +7,8 @@ jobs:
     strategy:
       matrix:
         ruby:
-          - 2.6
-          - 2.7
+          - 3.0
+          - 3.3
```
parkr, line 11: Quote these so they become strings when parsed rather than floats.

.travis.yml
```diff
@@ -1,8 +1,7 @@
 language: ruby
 rvm:
-  - 2.5
-  - 2.6
-  - 2.7
+  - 3.0
+  - 3.3
```
parkr, line 4: I think you can remove this Travis config file entirely.

parkr (approved):

parkr:
Ruby 3.3 support is added by https://github.com/github/pages-gem/pull/897. That needs to be merged first!

Dockerfile.alpine
```diff
@@ -1,4 +1,4 @@
-FROM ruby:2.7.3-alpine
+FROM ruby:3.3-alpine
```
parkr, line 1: I believe CI is attempting to pass the RUBY_VERSION build arg, but here it's always ruby 3.3.

Either remove the RUBY_VERSION arg from CI and script/cibuild-docker or support it here in Dockerfile.

parkr (approved):

lib/github-pages/dependencies.rb
```diff
@@ -63,7 +63,7 @@ def self.version_report
       require "nokogiri"
 
       {
-        "ruby" => RUBY_VERSION,
+        "ruby" => "3.3",
```
parkr, line 66: This will show unexpected values for non-Ruby 3.3 CI.

parkr (approved):

PR #895, opened 2023-12-20
Add Ruby 3.0, 3.1, 3.2, and 3.3 to CI, and retire 2.6

parkr:
Ruby is moving forward, and Pages must follow or risk security issues and poor developer experience.

PR #901, opened 2024-02-06
Bring back proper support for Ruby 3.x

yoannchaudet:
Address https://github.com/github/pages-gem/pull/895 on top of the latest changes in this repo.

Adding back all versions of Ruby still supported https://endoflife.date/ruby.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -7,8 +7,10 @@ jobs:
     strategy:
       matrix:
         ruby:
-          - '3.0'
-          - '3.3'
+          - 3.0
+          - 3.1
+          - 3.2
+          - 3.3
     steps:
     - uses: actions/checkout@v4
     - name: script/cibuild-docker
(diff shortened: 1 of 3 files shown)


.github/workflows/push.yml
```diff
@@ -7,8 +7,10 @@ jobs:
     strategy:
       matrix:
         ruby:
-          - '3.0'
-          - '3.3'
+          - 3.0
```
parkr, line 10: This one must be quoted, otherwise it's read as "3" and will use the latest Ruby 3.

parkr (changes requested):

.github/workflows/push.yml
yoannchaudet: Thanks for catching that and reminding me why I hate YAML 😁

.github/workflows/push.yml
```diff
@@ -7,8 +7,10 @@ jobs:
...
+          - 3.1
+          - 3.2
+          - 3.3
```
yoannchaudet, line 13:
```suggestion
          - '3.0'
          - '3.1'
          - '3.2'
          - '3.3'
```

tsusdere (approved):

PR #902, opened 2024-02-06
Bump dependencies (patch and minor only)

yoannchaudet:
Prepping a release so upgrading what we can (in theory) safely do for now.

### DIFF
--- a/lib/github-pages/dependencies.rb
+++ b/lib/github-pages/dependencies.rb
@@ -11,35 +11,35 @@ module GitHubPages
       "jekyll-sass-converter" => "1.5.2",
 
       # Converters
-      "kramdown" => "2.3.2",
+      "kramdown" => "2.4.0",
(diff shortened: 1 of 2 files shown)


lib/github-pages/dependencies.rb
```diff
@@ -11,35 +11,35 @@ class Dependencies
...
+      "rouge" => "3.30.0",
+      "github-pages-health-check" => "1.18.2",
 
       # Plugins
       "jekyll-redirect-from" => "0.16.0",
       "jekyll-sitemap" => "1.4.0",
-      "jekyll-feed" => "0.15.1",
-      "jekyll-gist" => "1.5.0",
+      "jekyll-feed" => "0.17.0",
+      "jekyll-gist" => "1.6.0",
```
yoannchaudet, line 27:
```suggestion
      "jekyll-gist" => "1.5.0",
```

PR #917, opened 2024-06-24
Upgrade to Jekyll 3.10.0

parkr:
No functionality is changed. Since Ruby 3.0 dropped webrick and Ruby 3.4 is going to drop csv, this adds them back as dependencies.

https://jekyllrb.com/news/2024/06/23/jekyll-3-10-0-released/

Fixes #899.
Fixes https://github.com/github/pages-gem/issues/887.
Fixes https://github.com/github/pages-gem/issues/864.
Fixes https://github.com/github/pages-gem/issues/752.

/cc @yoannchaudet

PR #914, opened 2024-03-22
Add webrick as Ruby 3 doesn’t include it

mrmanc:
Fixes #752. `webrick` wasn’t available it’s [not bundled as of Ruby v3.0.0](https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/). I’ve resolved this by adding it to the dependencies class.

PR #919, opened 2024-07-18
Parkr jekyll3.10

yoannchaudet:
Same as https://github.com/github/pages-gem/pull/917 but with an additional `jekyll-commonmark-ghpages` bump.

Also reconsider https://github.com/github/pages-gem/pull/914 by moving `webrick` to a gemspec dependency.

### DIFF
--- a/lib/github-pages/dependencies.rb
+++ b/lib/github-pages/dependencies.rb
@@ -9,7 +9,6 @@ module GitHubPages
       # Jekyll
       "jekyll" => "3.10.0",
       "jekyll-sass-converter" => "1.5.2",
-      "webrick" => "1.8.1",
 
       # Converters
       "kramdown" => "2.4.0",
--- a/lib/github-pages/dependencies.rb
+++ b/lib/github-pages/dependencies.rb
@@ -7,7 +7,7 @@ module GitHubPages
   class Dependencies
     VERSIONS = {
       # Jekyll
-      "jekyll" => "3.9.5",
+      "jekyll" => "3.10.0",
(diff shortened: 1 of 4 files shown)


lib/github-pages/dependencies.rb
```diff
@@ -7,14 +7,14 @@ module GitHubPages
...
-      "jekyll" => "3.9.5",
+      "jekyll" => "3.10.0",
       "jekyll-sass-converter" => "1.5.2",
       "webrick" => "1.8.1",
 
       # Converters
       "kramdown" => "2.4.0",
       "kramdown-parser-gfm" => "1.1.0",
-      "jekyll-commonmark-ghpages" => "0.4.0",
+      "jekyll-commonmark-ghpages" => "0.5.0",
```
yoannchaudet, line 17:
```suggestion
      "jekyll-commonmark-ghpages" => "0.5.1",
```

parkr:
Might be best to include Webrick in the gem spec directly? The way it is now folks will ask to upgrade it, but if it's in the gemspec directly then you can set it to be `~> X` and give users more freedom to upgrade as they wish.

yoannchaudet:
Instead of in the dependencies like in https://github.com/github/pages-gem/pull/914/files?

parkr:
> Instead of in the dependencies like in #914?

Yeah, that's what I was thinking. Anything in dependencies.rb is just more to manage without a lot of gain. It also doesn't matter for https://pages.github.com/versions/ since it's not used to build sites.

YiMysty (approved):

tsusdere (approved):

PR #921, opened 2024-08-05
Fix improperly bound regex

yoannchaudet:
This PR address a code vulnerability issue: https://github.com/github/pages-gem/security/code-scanning/1.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -16,7 +16,7 @@ RSpec.configure do |config|
 
   # Stub all GitHub API requests so they come back empty.
   config.before(:each) do
-    stub_request(:get, /api.github.com/)
+    stub_request(:get, /api\.github.com/)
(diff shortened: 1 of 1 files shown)


YiMysty (approved):

spec/spec_helper.rb
```diff
@@ -16,7 +16,7 @@
 
   # Stub all GitHub API requests so they come back empty.
   config.before(:each) do
-    stub_request(:get, /api.github.com/)
+    stub_request(:get, /api\.github.com/)
```
parkr, line 19: Should both periods be made literal?

parkr (approved):

spec/spec_helper.rb
yoannchaudet: yep I cannot read 😁. But I'll go over my other PR first, the one I dropped mid-way because it will help the CI!

